Issue #54, opened 2019-12-07
Add resample_draws


avehtari:
Add resample_draws function similar to thin_draws. In thin_draws there is 
```
sel_iterations <- seq(1, niterations, by = thin)
subset(x, iteration = sel_iterations)
```
resample_draws would get optional weight vector and resampling algorithm name, and `sel_iterations` would be determined by the resampling algorithm, e.g.
 - simple resampling without replacement
 - simple resampling with replacement
 - deterministic resampling
 - stratified resampling

Stratified resampling is currently used in rstan and rstanarm when importance sampling is used to improve advi or optimizing.

The reference for different resampling algorithms:
Kitagawa, G., Monte Carlo Filter and Smoother for Non-Gaussian Nonlinear State Space Models, Journal of Computational and Graphical Statistics, 5(1):1-25, 1996.

paul-buerkner:
Good idea! The only technical problem we will need to solve is that `subset` (or `subset_draws`) can only handle true subsetting so far, that is, they cannot select the same iteration twice as would be required for some resampling methods. We may need to add this possibility to `subset_draws` so that elements can be selected multiple times, at least when explicitely allowed by setting a flag accordingly. Or would it be better to add another method for this purpose?

MansMeg:
I could probably help out with this if you would need some help @paul-buerkner .

paul-buerkner:
Sure! Thank you! Let's discuss this on monday.

Måns Magnusson <notifications@github.com> schrieb am So., 8. Dez. 2019,
13:41:

avehtari:
Stratified resampling can be copied from rstan/rstanarm, and I can provide reference code and some documentation text for different resampling methods.

MansMeg:
Excellent! Can you give a pointer to the rstanarm code? It would be great to have it here in the issue.

avehtari:
`.sanple_indices` https://github.com/stan-dev/rstanarm/blob/master/R/stan_glm.fit.R#L1058
and how it is used https://github.com/stan-dev/rstanarm/blob/master/R/stan_glm.fit.R#L631

MansMeg:
Thanks!

paul-buerkner:
Non-unqiue subsetting is now possible by setting `unique = FALSE` in `subset_draws`. For example

```R
x <- example_draws()
# extract the first chain twice
subset_draws(x, chain = c(1, 1), unique = FALSE)
```

This feature should make implementing `resample_draws` possible @MansMeg

[...]

PR #65, opened 2020-04-01
support 'resample_draws'

paul-buerkner:
This PR implements #54. @MansMeg thank you for agreeing to review this PR!


NAMESPACE
```diff
@@ -89,6 +89,7 @@ S3method(repair_draws,draws_array)
 S3method(repair_draws,draws_df)
 S3method(repair_draws,draws_list)
 S3method(repair_draws,draws_matrix)
+S3method(resample_draws,draws)
```
MansMeg, line 92: Do we only want this methods to work with draws objects? Not draws_df, draws_list etc?

R/resample_draws.R
```diff
@@ -0,0 +1,176 @@
+#' Resample Draws Objects
+#'
+#' Resample [`draws`] objects according to given weights obtained,
+#' for instance, through importance sampling.
+#'
+#' @template args-methods-x
+#' @param weights A vector of positive weights of length equal to the
+#' number of draws in `draws` (as obtained by method [`ndraws`]).
+#' Draws with higher weights will be selected more often, usually
```
MansMeg, line 9: Is this sentence necessary? 
"Draws with higher weights will be selected more often, usually ... their weights."

R/resample_draws.R
```diff
@@ -0,0 +1,176 @@
...
+#' roughly proportional to their weights. How exactly the weights are handled
+#' depends on the `method` argument. Weights will be internally normalized
+#' if necessary.
```
MansMeg, line 12: Will they not always be normalized internally?

R/resample_draws.R
```diff
@@ -0,0 +1,176 @@
...
+#' @param method Name of the resampling method being applied.
+#' Possible choices are `"simple"` for simple random resampling with
+#' replacement, `"simple_no_replace"` for simple random resampling without
+#' replacement, `"stratified"` for stratified resampling with replacement
```
MansMeg, line 16: Stratified by what? I guess we - in the long run at least - need to specify what the different approaches are doing?

R/resample_draws.R
```diff
@@ -0,0 +1,176 @@
...
+resample_draws <- function(x, ...) {
+  UseMethod("resample_draws")
+}
+
+#' @rdname resample_draws
+#' @export
+resample_draws.draws <- function(x, weights, method = "stratified",
+                                 ndraws = NULL, ...) {
+  if (!allow_resample_draws(x)) {
+    stop2("'resample_draws' requires formats which can be subsetted ",
```
MansMeg, line 63: What is the reason for this?

R/resample_draws.R
```diff
@@ -0,0 +1,176 @@
...
+          "by individual draws, not only by iterations or chains.")
+  }
+  ndraws_total <- ndraws(x)
+  assert_numeric(weights, len = ndraws_total, lower = 0, null.ok = TRUE)
+  assert_choice(method, supported_resample_methods())
+  assert_number(ndraws, null.ok = TRUE, lower = 0)
+  weights <- weights / sum(weights)
+  method_fun <- paste0(".resample_", method)
+  method_fun <- get(method_fun, asNamespace("posterior"))
+  sel_draws <- method_fun(
```
MansMeg, line 73: maybe `draws_idx` or similar would be a more clear variable name? Now I expect a boolean vector (select/not select)?

NAMESPACE
paul-buerkner: `draws` is the parent class of all the formats so a `.draws` method will apply to all formats.

R/resample_draws.R
paul-buerkner: Probably not. I will remove it.

R/resample_draws.R
paul-buerkner: True. I will correct that.

R/resample_draws.R
paul-buerkner: This is a good point. I need to read the paper first :-D For now I just refer to the paper for details and we can change it later on.

R/resample_draws.R
paul-buerkner: After subsetting draws, the structure will no longer be compatible with a draws_array or draws_list object since, after subsetting, different chains will have different lengths. Do you see a way around this?

R/resample_draws.R
paul-buerkner: Good point. I will change that.

R/resample_draws.R
paul-buerkner: One option would be to drop the chain dimension (with a message). A similar thing technically happens `draws_df` objects already. I will work on this option.

paul-buerkner: Ok. Should be working now for all formats.

paul-buerkner:
Thanks @MansMeg for your review! I incorporated all the comments of yours. In particular, resampling should now be supported for all formats, via automatic merging of chains. Would you mind taking another look at the PR? I have also added you as contributor to posterior.

R/resample_draws.R
```diff
@@ -0,0 +1,176 @@
...
+    ndraws = ndraws,
+    ...
+  )
+  subset_draws(x, draw = sel_draws, unique = FALSE)
+}
+
+# simple random resampling with replacement
+# @return index vector of length 'ndraws'
+.resample_simple <- function(weights, ndraws, ...) {
+  ndraws <- default_ndraws_resample(ndraws, weights)
```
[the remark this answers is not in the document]
paul-buerkner: Done.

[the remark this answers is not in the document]
paul-buerkner: The reason I was initially handling this different was because `ndraws` default cannot be set for method `simple_no_replace`. I think the new way of handling this exception is now cleaner.

tests/testthat/test-resample_draws.R
```diff
@@ -0,0 +1,30 @@
+test_that("resample_draws returns expected format", {
```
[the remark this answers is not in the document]
paul-buerkner: Good idea. I have added such tests.

MansMeg:
Great! I do not think I have any more suggestions now. Good work @paul-buerkner!

paul-buerkner:
Thank you! I really appreciate your comments!

PR #76, opened 2020-06-11
try to fix issue #75

### LINKED ISSUE #75: weight_draws error
```r
x <- example_draws()
weights <- rexp(ndraws(x))
x <- weight_draws(x, weights = weights)

Error in dimnames(out) <- `*vtmp*` : 'dimnames' must be a list
```

paul-buerkner:
This PR should fix #75. @jgabry can you double check?

### DIFF
--- a/R/as_draws_array.R
+++ b/R/as_draws_array.R
@@ -126,7 +126,7 @@ draws_array <- function(..., .nchains = 1) {
   variables <- names(out)
   out <- unlist(out)
   out <- array(out, dim = c(niterations, .nchains, length(variables)))
-  dimnames(out)[[3]] <- variables
+  dimnames(out) <- list(NULL, NULL, variables)
   as_draws_array(out)
 }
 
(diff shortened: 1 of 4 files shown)


jgabry:
yeah will check now

jgabry:
It works, thanks! 

I just pushed a commit to your branch editing the examples a bit to demonstrate more clearly how the original weights can be recovered by using the different arguments to `weights.draws()`. In case you don't like my edits I'll hold off on merging this ;) Feel free to merge if you like my changes or after making more edits

jgabry:
Also, the reason I added those extra examples is because I got confused because the defaults value for the `log` argument of `weights.draws()` is different than `weights.psis()`. That could be an issue going forward...

paul-buerkner:
Thank you! These are really helpful doc edits! Merging this now.

paul-buerkner:
With regard to the `log = FALSE` default, I agree that the different default as compare to .psis is unfortunate but I think that most users expect standard normalized weights from a `weights` method, which should be reflected in the defaults. I am happy to get convinced otherwise of course.

PR #79, opened 2020-06-30
Add handling behaviour for NULL for *_ids(), variables() and n*()

### LINKED ISSUE #78: Is variables(NULL), niterations(NULL), etc. throwing intentional?
Example:
```
> posterior::variables(NULL)
Error in UseMethod("variables") : 
  no applicable method for 'variables' applied to an object of class "NULL"
```
Should it maybe return NULL instead silently, niterations(NULL) would return 0, etc? Or is this and intentional decision?

rok-cesnovar:
Fixes #78


paul-buerkner:
Thank you! Would you be ok with me moving the treatment of `NULL` outside of the generic and inside of a `*.NULL` method?

rok-cesnovar:
Done that in my first attempt but missed the export so got nowhere :). It should be fine now.

paul-buerkner:
Looks good to me! Thank you! Will merge this now.

paul-buerkner:
I have also added you as a contributor to the package.

rok-cesnovar:
That is very kind of you! Thanks!

jgabry:
Thanks @rok-cesnovar!

PR #86, opened 2020-09-02
Fix Blom 1958 reference

ahartikainen:
This PR fixes the reference.

This came up in https://github.com/arviz-devs/arviz/pull/1366#discussion_r479935719

Thanks @avehtari for the correction.


paul-buerkner:
Thanks!

Issue #88, opened 2020-09-21
.log_weight variable not hidden


n-kall:
It seems that unlike `.chain`, `.iteration` and `.draw`, `.log_weight` is not a hidden variable.
```
x <- example_draws()
weights <- rexp(ndraws(x))
x <- weight_draws(x, weights = weights)
variables(x)
```

```
 [1] "mu"          "tau"         "theta[1]"    "theta[2]"    "theta[3]"   
 [6] "theta[4]"    "theta[5]"    "theta[6]"    "theta[7]"    "theta[8]"   
[11] ".log_weight"
```
This leads to unexpected behavior such as counting it as a variable in `nvariables()` and summarising it with `summarise_draws()`.

paul-buerkner:
Yeah, that is a good point. I am not yet sure what the intended behavior should be to be honest as `.log_weight` has a somewhat different purpose than `.chain` etc.  @jgabry and @mjskay what is your opinion on hiding vs. not hiding `.log_weight`?

jgabry:
Yeah good question. This seems tricky because, like Paul was saying, `.log_weight` is a somewhat unique case. The chain, iteration, etc., are clearly metadata unlike the weights. But the weights also aren't regular variables. So they're somewhere in between. I think I would lean towards hiding them, but I don't have a strong opinion on it. 

What do others think?

paul-buerkner:
I would also tend to hide it. This also has the advantages that it won't be dropped automatically when subsetting draws after variables (without selecting `.log_weight`), which I think is desrible.

If we decide to go that route, we need to extend our meta-columns framework to apply to all draws formats not just to `draws_df` but that should be doable.

paul-buerkner:
Resolved via #91

PR #91, opened 2020-10-14
Support reserved variables in all formats

paul-buerkner:
This PR add supports for reserved variables in all formats. Right now, the only such reserved variable is `.log_weight` (#88) but we will certainly add more in the future. For example, certain HMC diagnostics such divergent transitions will probably be stored as such reserved variables. With the new functionality in place, making new variables reserved is a matter of writing their names once in a certain place.

There were quite a few changes necessary so I would ask like to ask @jgabry to review this PR.


R/draws-index.R
```diff
@@ -221,28 +231,10 @@ nvariables.NULL <- function(x) {
...
-}
-
-#' @export
-nvariables.draws_array <- function(x) {
-  dim(x)[3]
-}
-
-#' @export
-nvariables.draws_df <- function(x) {
+nvariables.draws <- function(x) {
```
jgabry, line 234: Should this also take a `reserved` argument and then pass it to `variables()`?

R/subset_draws.R
```diff
@@ -195,6 +160,114 @@ subset_dims <- function(x, ...) {
...
+      if (!is.null(chain)) {
+        x <- x[x$.chain %in% chain, ]
+      }
+      if (!is.null(iteration)) {
+        x <- x[x$.iteration %in% iteration, ]
+      }
+      x <- repair_draws(x, order = FALSE)
+    } else {
+      # non-unique subsetting is conceptually easier in 'draws_list' objects
+      # TODO: perform non-unique subsetting directly in the 'draws_df' object
```
jgabry, line 223: Instead of a TODO comment in the code should we open an issue for this?

man-roxygen/args-methods-reserved.R
```diff
@@ -0,0 +1,2 @@
+#' @param reserved Logical. Indicates whether reserved variables should be
+#' included in the output. Defaults to \code{FALSE}.
```
jgabry, line 2: Should we mention what the reserved variables are? I guess we'd have to remember to edit this when adding additional reserved variables, but it could be good to document them here. I'm OK either way.

R/remove_variables.R
```diff
@@ -0,0 +1,60 @@
+# remove variables from a draws object
+remove_variables <- function(x, ...) {
```
jgabry, line 2: The code for these methods is pretty simple, but maybe still worth adding some simple tests for these?

jgabry (commented):
This looks good! Just a few minor comments.

R/subset_draws.R
paul-buerkner: I made an issue (#92)

R/draws-index.R
paul-buerkner: Done.

man-roxygen/args-methods-reserved.R
paul-buerkner: Added `?'reserved-variables` doc page.

R/remove_variables.R
paul-buerkner: Added some unit tests.

paul-buerkner:
Thank you @jgabry for reviewing this PR and for your suggestions! I incoporated all of them and will merge this PR, after all checks have passed.

paul-buerkner:
Travis failed because something unrelated to posterior. Will merge anyway as all checks are passing locally.

jgabry:
Ok great, thanks!

PR #94, opened 2020-10-27
rstar calculation

ben18785:
- implements R* calculation (including uncertainty in statistic) via a single function
- added a split_chains_draws_df function: I realise this is basic and probably wants to be improved later to make it more widely applicable
- includes various tests


ben18785:
@paul-buerkner @jgabry at the moment the tests are failing (they work locally) because I suspect travis doesn't know to install caret. Any ideas?

rok-cesnovar:
The same error occurs on all tests: 

```
> test_check("posterior")
-- 1. Error: rstar returns reasonable values (@test-convergence.R#98)  ---------
package e1071 is required
Backtrace:
 1. posterior:::rstar(x)
 3. caret:::train.formula(...)
 5. caret:::train.default(x, y, weights = w, ...)
 6. caret:::evalSummaryFunction(...)
 7. ctrl$summaryFunction(testOutput, lev, method)
 8. caret::postResample(data[, "pred"], data[, "obs"])
 9. caret:::requireNamespaceQuietStop("e1071")
```

This is from appveyor. At least this error suggests that carret is available, but e1071 is not?

R/convergence.R
```diff
@@ -316,6 +316,144 @@ quantile2 <- function(x, probs = c(0.05, 0.95), names = TRUE, ...) {
...
+#'
+#' # with uncertainty, returns a vector of R* values
+#' hist(rstar(x, uncertainty = T))
+#' hist(rstar(x, uncertainty = T, nsim = 100))
+#'
+#' # can use other classification methods in caret library
+#' rstar(x, method = "knn")
+#'
+#' @export
+rstar <- function(x, split_chains=T, uncertainty=F, method=NULL, hyperparameters=NULL,
```
paul-buerkner, line 386: Please spell out TRUE and FALSE everywhere as T and F are not reserved in R and can be overwritten by the user.

R/convergence.R
```diff
@@ -316,6 +316,144 @@ quantile2 <- function(x, probs = c(0.05, 0.95), names = TRUE, ...) {
...
+                            training_proportion=0.7, nsim=1000, ...){
+
+  if(!requireNamespace("caret", quietly = T))
+    stop("Package \"caret\" needed for this function to work. Please install it.")
+  require(caret)
```
paul-buerkner, line 391: Would it be possible to access caret functions via `caret::` rather than calling `require(caret)`?

ben18785:
@rok-cesnovar thanks! Appveyor now seems to be failing to install a package called "digests". I'm not sure where this dependency comes from (I checked gbm, randomForest, caret and e1071 but it doesn't seem to depend on these)?

R/convergence.R
```diff
@@ -316,6 +316,144 @@ quantile2 <- function(x, probs = c(0.05, 0.95), names = TRUE, ...) {
...
+    caret_grid <- data.frame(interaction.depth=3,
+                         n.trees = 50,
+                         shrinkage=0.1,
+                         n.minobsinnode=10)
+  } else {
+    caret_grid <- hyperparameters
+  }
+
+  # remove iteration / draws columns and fit classifier
+  training_data <- training_data[, c(variables(training_data), ".chain")]
```
paul-buerkner, line 429: it might be more explicit to use `training_data$.iteration <- training_data$.draw <- NULL`

paul-buerkner, line 429: As removing .iteration and .draw invalidates a draws_df object, I recommend removing the draws classes from the object to avoid any unexpected behavior. This works, for example. via `class(training_data) <- class(training_data)[-(1:2)]

R/convergence.R
```diff
@@ -316,6 +316,144 @@ quantile2 <- function(x, probs = c(0.05, 0.95), names = TRUE, ...) {
...
+
+  if(training_proportion <= 0 || training_proportion >= 1)
+    stop("training_proportion must be greater than zero and less than 1.")
+
+  x <- as_draws_df(x)
+  if(split_chains)
+    x <- split_chains_draws_df(x)
+
+  # if only 1 param, add in a column of random noise
+  nparams <- nvariables(x)
```
paul-buerkner, line 405: could we rename the object `nparams` to `nvariables` for increase consistency across the posterior package code?

R/convergence.R
```diff
@@ -316,6 +316,144 @@ quantile2 <- function(x, probs = c(0.05, 0.95), names = TRUE, ...) {
...
+               tuneGrid = caret_grid,
+               ...)
+
+  # calculate classification accuracy then R*
+  nchains <- length(unique(testing_data$.chain))
+  if(uncertainty){
+    probs <- predict(object=fit, newdata=testing_data, type = "prob")
+    m_accuracy <- matrix(nrow = nrow(probs),
+                         ncol = nsim)
+    for(j in 1:nrow(probs)){
```
paul-buerkner, line 443: I recommend replacing `1:nrow(probs)` with `seq_len(NROW(probs))` as this is safer for edge cases.

R/convergence.R
```diff
@@ -570,3 +708,26 @@ fold_draws <- function(x) {
 should_return_NA <- function(x) {
   anyNA(x) || checkmate::anyInfinite(x) || is_constant(x)
 }
+
+# splits chains into halves within a `draws_df` object
```
paul-buerkner, line 712: Thank you for writing this method! I will make `split_chains` a separate generic but this can also be done after this PR has been merged.

tests/testthat/test-convergence.R
```diff
@@ -92,3 +92,95 @@ test_that("convergence diagnostics throw correct errors", {
   expect_error(ess_quantile(mu, probs = 1.2), "'probs' must contain values between 0 and 1")
   expect_error(mcse_quantile(mu, probs = 1.2), "'probs' must contain values between 0 and 1")
 })
+
```
paul-buerkner, line 95: Since rstar is such a different diagnostic from the others, I think it makes sense to move the corresponding tests into it's own file.

R/convergence.R
```diff
@@ -316,6 +316,144 @@ quantile2 <- function(x, probs = c(0.05, 0.95), names = TRUE, ...) {
   out
 }
 
+#' Calculate R star convergence diagnostic
```
paul-buerkner, line 319: Since rstar is such a different diagnostic than the others, I think it makes sense to move the code to a separate R file (called `rstar.R`).

.travis.yml
```diff
@@ -18,3 +18,8 @@ r_github_packages:
 after_success:
   - Rscript -e 'covr::codecov()'
 
+r_packages:
```
paul-buerkner, line 21: This should not be required if those packages are listed in the DESCRPITION.

R/convergence.R
paul-buerkner, line 386: Shall we make `rstar` generic with corresponding methods? @jgabry what do you think?

R/convergence.R
```diff
@@ -316,6 +316,144 @@ quantile2 <- function(x, probs = c(0.05, 0.95), names = TRUE, ...) {
...
+#' probability greater than chance. To do so, it fits a machine learning classifier
+#' to a training set of MCMC draws and evalutes its predictive accuracy on a testing
+#' set: giving the ratio of accuracy to predicting a chain uniformly at random.
+#'
+#' @param x a `draws_df` object or one coercible to a `draws_df` object.
+#'
+#' @param split_chains a Boolean indicating whether to split chains into two equal
+#' halves.
+#'
+#' @param uncertainty whether to provide a vector of R* values (if true) or a single
```
paul-buerkner, line 332: Can you explain here which uncertainty the vector contains.

R/convergence.R
```diff
@@ -316,6 +316,144 @@ quantile2 <- function(x, probs = c(0.05, 0.95), names = TRUE, ...) {
...
+#' value (if false).
+#'
+#' @param method machine learning classifer available in caret R package.
+#'
+#' @param hyperparameters hyperparameter settings for classifier given as a list.
+#'
+#' @param nsim number of R* values in returned vector if uncertainty is true.
```
paul-buerkner, line 339: I think it would be more in line with posterior naming conventitions to call the argument `nsimulations`

R/convergence.R
```diff
@@ -316,6 +316,144 @@ quantile2 <- function(x, probs = c(0.05, 0.95), names = TRUE, ...) {
...
+#'
+#' @param split_chains a Boolean indicating whether to split chains into two equal
+#' halves.
+#'
+#' @param uncertainty whether to provide a vector of R* values (if true) or a single
+#' value (if false).
+#'
+#' @param method machine learning classifer available in caret R package.
+#'
+#' @param hyperparameters hyperparameter settings for classifier given as a list.
```
paul-buerkner, line 337: I would say "named list" to make clear that element names are argument names.

R/convergence.R
```diff
@@ -316,6 +316,144 @@ quantile2 <- function(x, probs = c(0.05, 0.95), names = TRUE, ...) {
...
+#'
+#' @param training_proportion proportion of iterations used to train classifier.
+#'
+#' @details `rstar` provides a measure of MCMC convergence based on whether it is
+#' possible to determine the chain that generated a particular draw with a probability
+#' greater than chance. To do so, it fits a machine learning classifier to a subset of
+#' the original MCMC draws (the training set) and evaluates its predictive accuracy on
+#' the remaining draws (the testing set). If predictive accuracy exceeds chance (i.e.
+#' predicting the chain that generated a draw uniformly at random), the diagnostic
+#' measure, R* > 1, indicating that convergence has yet to occur.
```
paul-buerkner, line 349: Do we have guidlines to state how much above 1 ist problematic? It is ok if not but we should make this explicit.

R/convergence.R
```diff
@@ -316,6 +316,144 @@ quantile2 <- function(x, probs = c(0.05, 0.95), names = TRUE, ...) {
...
+#'
+#' @param x a `draws_df` object or one coercible to a `draws_df` object.
+#'
+#' @param split_chains a Boolean indicating whether to split chains into two equal
+#' halves.
+#'
+#' @param uncertainty whether to provide a vector of R* values (if true) or a single
+#' value (if false).
+#'
+#' @param method machine learning classifer available in caret R package.
```
paul-buerkner, line 335: Please make the default of `method` explicit.

paul-buerkner (changes requested):
Thank you for much for this PR! Is already looks really good and my comments and requests are all minor I think. Also, don't forget to add yourself as a contributor (role ctb) to the package in the DESCRIPTION.

R/convergence.R
ben18785: Done

R/convergence.R
ben18785: I've added a note on this since we don't have guidelines yet.

R/convergence.R
ben18785: Done.

R/convergence.R
ben18785: Done.

R/convergence.R
ben18785: Done.

R/convergence.R
ben18785: Happy to do so. I didn't want to here because that would then change nvariables from a (posterior) function to a value. Are you happy with that?

R/convergence.R
ben18785: Yep, good shout. Done.

R/convergence.R
ben18785: caret has some annoying features; one of them seems to be that, to access particular ML methods, the package itself needs to be loaded. For example, when I do:

`fit <- caret::train(.chain ~ .,
                      data = training_data,
                      method = "knn",
                      trControl = caret::trainControl(method = 'none'),
                      tuneGrid = caret_grid,
                      ...)`

I get an error message if caret isn't loaded. Whereas:

`
require(caret)
fit <- train(.chain ~ .,
                      data = training_data,
                      method = "knn",
                      trControl = trainControl(method = 'none'),
                      tuneGrid = caret_grid,
                      ...)`

works. What do you think?

R/convergence.R
ben18785: Done. (Didn't know that -- good to know!)

R/convergence.R
ben18785: Done.

R/convergence.R
ben18785: Great.

R/convergence.R
ben18785: Done.

.travis.yml
ben18785: I was getting a failure from Travis before about it not finding caret. I thought it was because this wasn't being installed because it's only in `Suggests` rather than `Imports`. That's why I put it in there...

tests/testthat/test-convergence.R
ben18785: Done.

R/convergence.R
ben18785: Done.

ben18785:
I'm getting an error from appveyor for testing the `resample_draws` function. It seems to be not throwing an error (when the test says is should)?

rok-cesnovar:
The resample_draws error is also occurring on master, so that is not the fault of this PR.

paul-buerkner:
I am going to look at this error separately

Rok Češnovar <notifications@github.com> schrieb am Di., 27. Okt. 2020,
17:59:

ben18785:
@paul-buerkner @rok-cesnovar thanks. Will ignore then.

R/convergence.R
paul-buerkner: I see. I have no experience with caret myself and I would think that `require(caret)` is ok in this case if we cannot use `::` reliably.

R/convergence.R
paul-buerkner: The scoping of vectors and functions is different in R so that `nvariables(...)` still calls the function even if another non-function object in scope has the same name. Alternatively, we could call the value `nvars`. Both options are fine for me. I just don't like the `nparams` since calling things 'variables' instead of 'parameters' was a design decision in posterior.

paul-buerkner:
I hope I fixed the error on master.

Am Di., 27. Okt. 2020 um 18:01 Uhr schrieb ben18785 <
notifications@github.com>:

R/rstar.R
```diff
@@ -0,0 +1,174 @@
+# Copyright (C) 2012, 2013, 2014, 2015, 2016, 2017, 2018 Trustees of Columbia University
```
paul-buerkner, line 1: I am not sure this copyright is correct. Shouldn't it rather be you and Aki 2020?

R/rstar.R
```diff
@@ -0,0 +1,174 @@
...
+#' @examples
+#' x <- example_draws("eight_schools")
+#' rstar(x)
+#' rstar(x, split_chains = F)
+#' rstar(x, method = "gbm")
+#' # can pass additional arguments to methods
+#' rstar(x, method = "gbm", verbose = F)
+#'
+#' # with uncertainty, returns a vector of R* values
+#' hist(rstar(x, uncertainty = T))
```
paul-buerkner, line 71: Can you fix T and F in the doc as well?

ben18785: Done.

R/convergence.R
ben18785: Cool. Done.

R/convergence.R
ben18785: Thanks. Ok -- will keep as is.

R/rstar.R
ben18785: Ok, I've changed it to:

# Copyright (C) 2012, 2013, 2014, 2015, 2016, 2017, 2018 Trustees of Columbia University
# Copyright (C) 2020 Ben Lambert, Aki Vehtari

Let me know if this isn't ok, as I don't have much experience with these sorts of thing.

paul-buerkner:
The Columbia copyright was because we took some of the convergence
diagnostics code from rstan but this is not the case for rstar so I think
the Columbia copyright can be removed.

ben18785 <notifications@github.com> schrieb am Di., 27. Okt. 2020, 18:58:

ben18785:
> The Columbia copyright was because we took some of the convergence diagnostics code from rstan but this is not the case for rstar so I think the Columbia copyright can be removed. ben18785 <notifications@github.com> schrieb am Di., 27. Okt. 2020,

Ok, have removed it.

paul-buerkner (approved):

paul-buerkner:
I am happy with the PR! Thank you @ben18785 for the quick handling of my requests!

@avehtari do you see anything that needs changing before merging this PR?

R/convergence.R
jgabry: I might be wrong but I would think using `require()` inside package code would be discouraged by CRAN. I know that using `library()` is either not allowed or strongly discouraged, so I would think they would similarly discourage using `require()`. The reason, I assume, is that calling a function in one package shouldn't result in another package being loaded because that affects the global environment. That is, once we call `require(caret)` then caret is loaded for the user even after the `rstar()` function is exited, which can result in name clashes and other unexpected behavior.

jgabry:
Thanks @ben18785! And thanks @paul-buerkner for the review. The only issue I see is the comment I made about using `require()` in package code, otherwise this looks great.

R/convergence.R
paul-buerkner: Yeah. You are right. This is what CRAN says (https://cran.r-project.org/doc/manuals/r-release/R-exts.html):

"R code in the package should call library or require only exceptionally. Such calls are never needed for packages listed in ‘Depends’ as they will already be on the search path. It used to be common practice to use require calls for packages listed in ‘Suggests’ in functions which used their functionality, but nowadays it is better to access such functionality via :: calls. "

We could 

1) state in the comments to CRAN that the `require` call is necessary to make caret work. 
2) figure out a workaround which works with `::` calls (not sure how though). 
3) file an issue for caret to fix its use when not loaded and hope it will be fixed soon.

R/rstar.R
```diff
@@ -0,0 +1,193 @@
...
+#'
+#' # can use other classification methods in caret library
+#' rstar(x, method = "knn")
+#'
+#' @export
+rstar <-
+  function(x,
+           split_chains = TRUE,
+           uncertainty = FALSE,
+           method = NULL,
```
jgabry, line 89: Is there a reason for `method=NULL` and not `method="rf"`?

R/convergence.R
jgabry: What if we check if caret is loaded (on the search path?) and if not ask the user to manually load it? Would that work?

jgabry:
Right now R CMD check shows the following: 

```
 checking dependencies in R code ... NOTE
  'library' or 'require' call to ‘caret’ in package code.
    Please use :: or requireNamespace() instead.
    See section 'Suggested packages' in the 'Writing R Extensions' manual.

❯ checking R code for possible problems ... NOTE
  rstar: no visible global function definition for ‘train’
  rstar: no visible global function definition for ‘trainControl’
  Undefined global functions or variables:
    train trainControl
```

The former is related to my comment about `require()`. For the latter, I think prefixing these functions with `caret::` is required.

paul-buerkner:
I think that would work and sounds like a good approach to me.

Am Di., 27. Okt. 2020 um 20:05 Uhr schrieb Jonah Gabry <
notifications@github.com>:

R/rstar.R
```diff
@@ -0,0 +1,197 @@
...
+#'   For lower dimensional targets, gradient boosted models (called via
+#'   `method="gbm"`) tend to have a higher classification accuracy. On a given
+#'   MCMC sample, it is recommended to try both of these classifiers.
+#'
+#' @return A numeric vector of length 1 (by default) or length `nsimulations`
+#'   (if `uncertainty=TRUE`).
+#'
+#' @references Ben Lambert, Aki Vehtari (2020) R*: A robust MCMC convergence
+#'   diagnostic with uncertainty using gradient-boosted machines \emph{arXiv
+#'   preprint} \code{arXiv:TBD}.
```
jgabry, line 69: When do you plan on submitting to arXiv? If this isn't on arXiv yet (I'm assuming since it says TBD) is there somewhere else people can find the draft? I'm a bit hesitant to add a reference to a paper until it's available to read somewhere.

ben18785: Good spot. I just added the arxiv link.

R/rstar.R
ben18785: Nope. Changed.

R/rstar.R
ben18785: done

R/convergence.R
ben18785: @jgabry I like your idea. Will try to do this later tonight.

jgabry: Thanks!

ben18785: I've changed this now but it's failing because Github actions isn't finding the caret, gbm, randomForest and e1071 packages. These are suggests which previously I was requiring in Travis. Any idea how to handle these in Github actions?

jgabry: Not 100% sure but I think the place to do it is probably here: 

https://github.com/stan-dev/posterior/blob/cfa6dc5f04195d6b450270818ccb04a858f7bfec/.github/workflows/rcmdcheck.yml#L74-L77

The `remotes::install_cran()` function can take a vector of package names.

paul-buerkner:
Thank you @jgabry for your additional comments and help and thank you @ben18785 for incorporating them so quickly. Before we merge, @jgabry is there anything you think needs to be changed?

jgabry (approved):
Looks great!

Issue #93, opened 2020-10-22
Add Split Chains method


paul-buerkner:
We should add a `split_chains` method for use in the new R* diagnostic (#87) and perhaps also in other methods going forward.

paul-buerkner:
This has been done already, just forgot to close this issue.

PR #96, opened 2020-10-29
Implement 'split_chains'

paul-buerkner:
This PR implements the `split_chains' method (#93). @jgabry would you be willing to review this PR?


jgabry:
I just pushed one commit to fix a failing test, but otherwise I think this looks good. 

One quick question: is the reason we have the separate `.split_chains()` internal function just that the one variable matrices used by the diagnostics aren't proper draws objects?

paul-buerkner:
Thank you! Yes .split_chains exists for the diagnostics which require a
simple iteration x chain matrix.

Jonah Gabry <notifications@github.com> schrieb am Do., 29. Okt. 2020, 21:51:

jgabry (approved):

Issue #98, opened 2020-11-13
Summarise draws is probibitively slow for large number of parameters


spinkney:
I updated `cmdstanr` to use `data.table` and then noticed that the second bottleneck is the `posterior` package. For small number of parameters `data.table` isn't much different but once the number of parameters grows to 10s or 100s of thousands the `summarise_draws()` function becomes much, much slower.

For example, here is a benchmark of 40901 parameter summary using mean, median, and mad of the posterior `summarise_draws` vs a custom function using `data.table`. The difference is 5x slower for posterior version.
```r
# 40901 parameters
> mbm <- microbenchmark(
+   ex_posterior_summary = summarise_draws(ex_posterior, mean, median, mad),
+   ex_dt_summary = summarise_draws_dt(ex_posterior),
+   times = 10L
+ )
> mbm
Unit: seconds
                 expr       min        lq      mean    median        uq       max neval
 ex_posterior_summary 174.41825 175.31964 178.12328 177.51151 180.00368 185.84406    10
        ex_dt_summary  33.50971  33.85208  34.53455  34.47042  34.86485  35.95574    10
> 178/34
[1] 5.235294
```
When I increase the parameters to 79436. The difference is now 6.67x slower. 
```r
dim(ex_posterior)
[1]  1000 79436
> mbm
Unit: seconds
                 expr       min        lq      mean    median        uq       max neval
 ex_posterior_summary 541.82929 541.82929 541.82929 541.82929 541.82929 541.82929     1
        ex_dt_summary  81.64224  81.64224  81.64224  81.64224  81.64224  81.64224     1
> 541/81
[1] 6.679012
```
The function `summarise_draws_dt` I created using `data.table` is
```r
summarise_draws_dt <-function(d) {
    ex_dt <- as.data.table(d)
    dt_sum <- t(ex_dt[, as.list(unlist(lapply(.SD, function(x) list(mean = mean(x), 
                                                                    median = median(x),
                                                                    # sd = sd(x),
                                                                    mad = mad(x)
                                                                ))))])
    dt_sum <- as.data.table(dt_sum, keep.rownames = T)
    dt_sum[, c("variable", "cast") := tstrsplit(rn, ".", fixed=TRUE)]
    return(dcast(dt_sum, variable ~ cast, value.var = "V1", fun.aggregate = mean))
}
````

Equivalency of output
```r
# output is equivalent
# > summarise_draws
# A tibble: 14 x 4
#   variable           mean  median     mad
[...]

PR #99, opened 2020-11-13
Special case as_draws_array for length 1

rok-cesnovar:
When we use as_draws_array and we only have 1 chain we could optimize the function a bit. 

Testing scenario: 
```r
library(cmdstanr)
model_code <- "
parameters {
  real k;
}

transformed parameters {
  real x[50000] = rep_array(k, 50000);
}

model {
  k ~ normal(0, 1);
}
"
model_file <- write_stan_file(model_code)
mod <- cmdstan_model(model_file)
fit <- mod$sample(iter_sampling = 2000, validate_csv = FALSE, chains = 1)

data <- data.table::fread(cmd= paste0("grep -v '^#' ", fit$output_files()[1]))

times <- c()
for (i in 1:10) {
  start <- Sys.time()
  d_a <- posterior::as_draws_array(data)
  times <- c(times, as.double((Sys.time()-start), units = "secs"))
}
print(times)
```

this branch:
```
1.838615 1.740373 1.687281 1.830399 1.704403 1.841837 1.646028 1.733209 1.632276 1.731928
```
mean 1.74s

master:
```
2.565165 2.900125 2.978846 2.799916 2.859506 3.029445 2.660265 2.807306 2.583319 2.838322
```
mean 2.8s

So roughly 40% faster. 

This is important for cmdstanr where we read results one chain at a time. But probably makes sense anyway.
I am looking into other bottlenecks so if you prefer to review in larger chunks we can keep the PR open and I can note when I am done.

### DIFF
--- a/R/as_draws_array.R
+++ b/R/as_draws_array.R
@@ -60,7 +60,7 @@ as_draws_array.draws_df <- function(x, ...) {
     if (length(chains) == 1) {
       out[[i]] <- x
     } else {
-      out[[i]] <- x[x$.chain == i, ]      
+      out[[i]] <- x[x$.chain == i, ]
     }
     out[[i]] <- remove_reserved_df_variables(out[[i]])
     out[[i]] <- as.matrix(out[[i]])
@@ -161,7 +161,14 @@ is_draws_array_like <- function(x) {
 # convert a list of matrices to an array
 as_array_matrix_list <- function(x) {
   stopifnot(is.list(x))
-  x <- abind::abind(x, along = 3L)
+  if (length(x) == 1) {
+    tmp <- dimnames(x[[1]])
+    x <- x[[1]]
+    dim(x) <- c(dim(x), 1)
+    dimnames(x) <- tmp
+  } else {
+    x <- abind::abind(x, along = 3L)
+  }
   x <- aperm(x, c(1, 3, 2))
 }
 
@@ -183,3 +190,4 @@ empty_draws_array <- function(variables = character(0), nchains = 0,
   class(out) <- class_draws_array()
   out
 }
+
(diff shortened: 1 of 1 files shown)


R/as_draws_array.R
```diff
@@ -57,7 +57,11 @@ as_draws_array.draws_df <- function(x, ...) {
   chains <- chain_ids(x)
   out <- vector("list", length(chains))
   for (i in seq_along(out)) {
-    out[[i]] <- x[x$.chain == i, ]
+    if (length(chains) == 1) {
+      out[[i]] <- x
+    } else {
+      out[[i]] <- x[x$.chain == i, ]      
+    }
```
rok-cesnovar, line 64: The biggest bottleneck now is the as.matrix call below, but that probably cant be avoided.

paul-buerkner: Indeed. We have to transform a list-like object (columns are elements of a list) into a numeric matrix in column major order. I am not sure we can do this faster than base R here.

paul-buerkner:
Thank you! I am fine with both, keeping the PR open for more changes or merging just this fix. What is your preference?

R/as_draws_array.R
rok-cesnovar: Yeah, I figured.

rok-cesnovar:
I think this one is worth it on its own. Will open subsequent ones if something worthy comes up.

paul-buerkner (approved):
The PR looks good to me. Thank you! @jgabry do you want to take a look before we merge?

jgabry:
Thanks! It looks good to me too. Should we bump the version number when we merge this? That way we can require this version for cmdstanr.

rok-cesnovar:
Lets hold off for a few more days with the bump, I might have time to look into other bottlnecks causing #98 
If I dont find any till next week, we can then bump.

jgabry:
Ok, sounds good thanks!

jgabry:
Since @paul-buerkner already approve this I'll go ahead and merge now and then we can do separate PRs for other fixes (e.g., if you or someone else figures out #98)

PR #102, opened 2020-11-27
allow drop=TRUE for draws_array

jgabry:
Allows drop=TRUE when indexing draws_array objects as discussed in https://github.com/stan-dev/loo/pull/161.

### DIFF
--- a/R/as_draws_array.R
+++ b/R/as_draws_array.R
@@ -154,7 +154,7 @@ is_draws_array_like <- function(x) {
   # TODO: allow for argument 'reserved' as in '[.draws_df'
   #   right now this fails because NextMethod() cannot ignore arguments
   out <- NextMethod("[", drop = drop)
-  if (length(dim(out)) == 3) {
+  if (length(dim(out)) == length(dim(x))) {
     class(out) <- class(x)
   }
   out
--- a/R/as_draws_array.R
+++ b/R/as_draws_array.R
@@ -150,10 +150,10 @@ is_draws_array_like <- function(x) {
 
 #' @export
 `[.draws_array` <- function(x, i, j, ..., drop = FALSE) {
-  # TODO: add a warning that 'drop' is ignored?
+  # TODO: add a warning that 'drop' can lead to also dropping the class?
   # TODO: allow for argument 'reserved' as in '[.draws_df'
   #   right now this fails because NextMethod() cannot ignore arguments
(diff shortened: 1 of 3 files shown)


jgabry:
I guess this isn't as simple as I thought. Seems to cause a lot of test failures.

paul-buerkner:
I will take a look tomorrow but note that we need to drop the class manually if we see the dimension to be in fact reduced.

R/as_draws_array.R
```diff
@@ -150,10 +150,10 @@ is_draws_array_like <- function(x) {
 
 #' @export
 `[.draws_array` <- function(x, i, j, ..., drop = FALSE) {
-  # TODO: add a warning that 'drop' is ignored?
+  # TODO: add a warning that 'drop' can lead to also dropping the class?
   # TODO: allow for argument 'reserved' as in '[.draws_df'
   #   right now this fails because NextMethod() cannot ignore arguments
-  out <- NextMethod("[", drop = FALSE)
+  out <- NextMethod("[")
   class(out) <- class(x)
```
jgabry, line 157: @paul-buerkner I guess the reason this doesn't work yet is because it's still setting the class to draws_array before exiting the function. However, the logic you used to get around this with draws_df doesn't quite work here since we don't have the "reserved" argument. Ideas?

jgabry:
Ok thanks, I just noticed that too and made a comment about it. I'll see if I can fix that and you can take a look tomorrow or whenever you have a chance.

jgabry:
Ok I think I figured that out. At least all the tests are passing now. I also added a new test. Wasn't sure where to put it but since it involved indexing I put it in test-draws-index.R (we can move it if you want).

paul-buerkner:
The PR looks good. Thank you @jgabry. Shall we use the same approach for `draws_matrix` objects? It would seem consistent for me to do so. And if yes, shall we add it to the same PR?

paul-buerkner (approved):
Approved but seem my comment about `draws_matrix` objects above.

jgabry:
Yeah it makes sense to do the same with matrices. I can add that to this PR this week.

jgabry:
@paul-buerkner I just added the same thing for draws_matrix. Do you want to take a look before we merge?

paul-buerkner:
Looks good to me. Thank you! Merging now.

PR #116, opened 2021-03-24
allow roundtrip draws_df -> data.frame -> draws_df, closes #114

### LINKED ISSUE #114: can't easily roundtrip draws_df -> data frame -> draws_df
Something I discovered while looking at #112 is that you can't currently roundtrip to a data frame (or tibble) and back to a draws_df easily due to the `.draw` column. E.g. this throws an error:

```r
> as_draws_df(as_tibble(as_draws_df(example_draws())))
 Error: Variable names {'.draw'} are reserved. 
```

When I would expect it should work.

This makes some workflows with data frames a bit more cumbersome due to the need to drop `.draw` if transformations of the data frame cause the draws_df class to be dropped. Since it is recalculated anyway, one easy fix would be to drop the `.draw` column before this check here:

https://github.com/stan-dev/posterior/blob/321644904edf0f4dc9f82ece32e8c3bf91fd4089/R/as_draws_df.R#L174

If that seems sensible I'll add a test and a fix.

mjskay:


paul-buerkner:
Thanks!

Issue #117, opened 2021-03-26
Testing the new `rvar` format in practice


paul-buerkner:
The recently merged `rvar` feature (#106) is a really big step forward for posterior (thank you, @mjskay)!

Due to its complexity, I expect minor problems to be popping up here and there that we should aim to iron out before the CRAN release. If you find some problems with the new format, please post them here in this issue.

paul-buerkner:
I start with a few minor discussion points:

- I see that one file is called `rvar-.R` I wonder if the `-` is intentional? Otherwise, I would rename this file.

- in `rvar()`, it seems that the `dim` argument cannot alter the first as `dim` is only used after conversion to `rvar` which suppresses the first array dimension because this is where we store draws. I agree with this behavior but this is likely to be unintuitive to R users, so it should be clearly documented.

- I think the doc pages of `rvar()` and `as_rvar()` could use some simple examples.

bbbales2:
In the vignette there seems to be a missing link:

```
do; see **TODO: INSERT LINK** for an introduction
```

Very excited about this! Thanks all

mjskay:
Thanks @paul-buerkner for opening a place to collect feedback on `rvar` stuff. I'm excited for folks to kick the tires on this!

> I see that one file is called `rvar-.R` I wonder if the `-` is intentional? Otherwise, I would rename this file.

It's a convention I've seen in other packages that keeps `rvar-.R` at the top of the directory sort order relative to other `rvar-...R` files, which I find nice when scanning through the file list. But I don't feel strongly about it, so if you don't like the convention I'll drop it. Let me know.

Re: the other two issues, adding checkboxes so I don't forget:

- [ ] betters docs for `dim` argument to `rvar()`
- [ ] simple examples for `rvar()` and `as_rvar()`

@bbbales2 thanks for reminding me, I forgot about that stub! Since that can't be filled in until the draws vignette exists, I'll add a reminder to do that in #42.

Adding one more thought: is `draws_of()` the right name for that function? The more I think about it, the function is mis-named given that it does not return a `draws` object. Perhaps `array_of()` would be better?

paul-buerkner:
Ah I remember, you have explained this `rvar-.R` to me at some point but I forgot. Thank you. 

I will have to think about the `draws_of()` idea a little more.

avehtari:
Reporting some inconsistent behavior. mean, median, quantile ok, max and min wrong result, sd error.
```
[...]

PR #123, opened 2021-04-09
Reorganize and expand rvar summary functions, for #117

mjskay:
This PR makes a number of changes towards a more consistent naming scheme for summary functions for rvars, as discussed in #117. 

A summary of changes:

- Splits summary functions into summaries **within** draws (prefixed `rvar_`, and which return `rvar` objects) and summaries **over** draws (no prefix, and which return arrays). A description of these functions can be found in the rewritten "Expectations and summary functions" section of vignette/rvar.Rmd; in particular see the table at the end of that section.
- Makes `sd()`, `var()`, and `mad()` generic in order to support implementations of those summary functions for `rvar`s.
- Adds an `rvar_apply()` function that acts like base `apply()` but simplifies results down to an `rvar` array and uses `rvar`-style broadcasting. See the new "Applying functions over rvars" section in vignette/rvar.Rmd.

I have not made existing diagnostic functions generic nor have I added rvar implementations for them in this PR. I decided this one is big enough and will leave that to a future PR.

Since this is a bigger PR, @paul-buerkner I would appreciate your review and thoughts. @avehtari since your testing prompted a number of these changes, your feedback would also be appreciated. Thanks!


R/rvar-.R
```diff
@@ -546,10 +546,17 @@ copy_dimnames <- function(src, src_i, dst, dst_i) {
 # helpers: applying functions over rvars ----------------------------------
 
 # apply a summary function within each draw of the rvar (dropping other dimensions)
-summarise_rvar_within_draws <- function(x, .f, ...) {
+summarise_rvar_within_draws <- function(x, .f, ..., .transpose = FALSE, .when_empty = .f()) {
```
paul-buerkner, line 549: is it safe to call all kinds of `.f` without input? Perhaps set `.when_empty` to a safer default?

R/rvar-.R
```diff
@@ -546,10 +546,17 @@ copy_dimnames <- function(src, src_i, dst, dst_i) {
...
   draws <- draws_of(x)
   dim <- dim(draws)
-  new_rvar(apply(draws, 1, .f, ...), .nchains = nchains(x))
```
paul-buerkner, line 552: I think that at least `.transpose` requires an input check via `as_one_logical`

R/rvar-summaries-over-draws.R
```diff
@@ -0,0 +1,187 @@
...
+#' the case of `range()`, see 'Details').
+#'
+#' @param x an [`rvar`]
+#' @param ... further arguments passed to underlying functions (e.g., `base::mean()`
+#' or `base::median()`), such as `na.rm`.
+#'
+#' @details
+#'
+#' Summaries include expectations (`E()` or `mean()`), probabilities (`Pr()`),
+#' medians (`median()`), spread (`variance()`, `sd()`, `mad()`), sums and
```
paul-buerkner, line 16: with the latest changes, can we add `var` to the list as well?

R/rvar-summaries-within-draws.R
```diff
@@ -0,0 +1,160 @@
...
+
+#' Summaries of random variables over array elements, within draws
+#'
+#' Compute summaries of random variables over array elements and within draws,
+#' producing a new random variable of length 1 (except in the case of
+#' `rvar_range()`, see 'Details').
+#'
+#' @param ... [`rvar`]s
+#' @template args-rvar-summaries-na.rm
+#' @param constant For `rvar_mad()`, a scale factor for computing the median
```
paul-buerkner, line 11: refer to `stats::mad` for justification of this constant default?

R/rvar-summaries-within-draws.R
```diff
@@ -0,0 +1,160 @@
...
+#'
+#' set.seed(5678)
+#' x = rvar_rng(rnorm, 4, mean = 1:4, sd = 2)
+#'
+#' # These will give similar results to mean(1:4),
+#' # median(1:4), sum(1:4), prod(1:4), etc
+#' rvar_mean(x)
+#' rvar_median(x)
+#' rvar_sum(x)
+#' rvar_prod(x)
```
paul-buerkner, line 42: add `rvar_range` example to illustrate the special case output?

R/rvar-summaries-within-draws.R
```diff
@@ -0,0 +1,160 @@
...
+rvar_mad <- function(..., constant = 1.4826, na.rm = FALSE, low = FALSE, high = FALSE) {
+  summarise_rvar_within_draws(c(...), mad, constant = constant, na.rm = na.rm, low = low, high = high)
+}
+
+
+# range -------------------------------------------------------------------
+
+#' @rdname rvar-summaries-within-draws
+#' @export
+rvar_range <- function(..., na.rm = FALSE) summarise_rvar_within_draws(c(...), range, na.rm = na.rm, .transpose = TRUE)
```
paul-buerkner, line 100: since we know how to handle range already with more than 1 ouput, could it make sense to also add `rvar_quantile`?

vignettes/rvar.Rmd
```diff
@@ -151,28 +151,63 @@ includes:
...
+
+1. Summary functions that mimic base-R vector summary functions, except applied to `rvar` vectors. These apply their summaries **over** elements of the input vectors **within** each draw, generally returning an `rvar` of length 1. These functions are prefixed with `rvar_` as a reminder that they return `rvar`s. Here is an example of `rvar_mean()`:
+  
+    ```{r rvar_mean_mu}
+    rvar_mean(mu)
+    ```
+
+2. Summary functions that summarise **within** elements of input vectors and **over** draws. These summary functions generally return base arrays (`numeric` or `logical`) of the same shape as the input `rvar`, and are especially useful for diagnostic summaries. These summary functions are not prefixed with `rvar_` as they do not return `rvar`s. Here is an example of `mean()`:
+
+    ```{r mean_mu}
```
paul-buerkner, line 187: There seem to be a few leading whitespaces here that could be removed.

paul-buerkner (changes requested):
Very well done! I continue to be impressed by the magic you created via `rvar`s. I just have a few minor comments that should not be difficult to address.

avehtari:
A quick note that, at least so far porting my motorcycle example to rvars seems to be working and behavior of rvar is as I would expect. Works consistently also for optimizing result (one chain, one iteration)

R/rvar-.R
mjskay: the intent here is that the `rvar_` versions of functions should yield similar results (or similar errors) to base functions given empty input. So e.g.:

```
> sum()
[1] 0
> rvar_sum()
rvar<1>[1] mean ± sd:
[1] 0 ± NA 
> mean()
Error in mean.default() : argument "x" is missing, with no default
> rvar_mean()
Error in mean.default() : argument "x" is missing, with no default 
```

That said, in some cases the default error message is a little obscure; for example:

```
> sd()
 Error in is.data.frame(x) : argument "x" is missing, with no default 
> rvar_sd()
 Error in is.data.frame(x) : argument "x" is missing, with no default 
```

Throwing an error is still the correct behavior, but I think we can probably provide a better error message via `.when_empty`

R/rvar-.R
mjskay: summarise_rvar_within_draws() is only called from internal code, and `.transpose` is not exposed to user input, so it should not be possible for it to be anything but a single logical value (in fact it's only used to implement rvar_range at the moment). So I would not expect the input check to ever fail. That said, if you prefer having the check there I will add it.

R/rvar-summaries-over-draws.R
mjskay: good point!

R/rvar-summaries-within-draws.R
mjskay: good idea

vignettes/rvar.Rmd
mjskay: The leading whitespace here ensures that the code block is nested under the list item above when rendered to HTML. It's a bit odd I admit but seems to be the correct way to do that (it must be exactly 4 spaces)

R/rvar-summaries-within-draws.R
mjskay: Good idea. I suppose we could do a quantile2 implementation as well? I've added them to the list in #117 (https://github.com/stan-dev/posterior/issues/117#issuecomment-811484159) and will handle them in the next round of changes (not this PR).

R/rvar-summaries-within-draws.R
mjskay: ah good point, done

R/rvar-.R
mjskay: I've updated it so that the functions that should throw errors now have more sensible error messages. Specifically `rvar_mean()`, `rvar_median()`, `rvar_sd()`, `rvar_var()`, and `rvar_mad()` now follow this pattern:

```
> rvar_median()
 Error: in rvar_median(): cannot take median of empty vector 
```

Does that sound good?

mjskay:
> Very well done! I continue to be impressed by the magic you created via rvars. I just have a few minor comments that should not be difficult to address.

Thanks @paul-buerkner! Apologies for the delay, the quarter is still going here :). I addressed almost everything but left two minor questions above when you get the chance.

> A quick note that, at least so far porting my motorcycle example to rvars seems to be working and behavior of rvar is as I would expect. Works consistently also for optimizing result (one chain, one iteration)

Awesome! Thanks again for putting the API through its paces.

R/rvar-.R
paul-buerkner: Sounds good to me.

R/rvar-.R
paul-buerkner: Thanks, that make sense and is fine as is.

paul-buerkner (approved):

paul-buerkner:
Thanks! Looks good now. Anything you want to change before we merge?

mjskay:
Nah, let's merge this one.

paul-buerkner:
Very well :-)

R/rvar-.R
jgabry: Nice, these error messages are much better than the ones in base R

Issue #92, opened 2020-10-16
Perform non-unique subsetting directly in the 'draws_df' object


paul-buerkner:
Right now, non-unique subsetting for `draws_df` objects via `subset_draws(..., unique = FALSE)` is done by first transforming to the `draws_list` format, subsetting, and transforming it back to `draws_df`. This may create a lot of run time overhead for big models and could be avoided by directly doing non-unique subsetting in `draws_df`.

jgabry:
fixed in #128

PR #128, opened 2021-05-12
perform non-unique subsetting directly in draws_df

Ozan147:
Hey @jgabry, this suggestion addresses #92.


jgabry:
Thanks @Ozan147!

paul-buerkner:
Looks good to me, thanks! Do we already have enough tests for this case in place? If yes, please feel free to merge.

jgabry:
I just added a test that compares non unique subsetting for draws_df and draws_list. If CI passes then I think we can go ahead and merge this.

Ozan147:
@jgabry, only row names end up being different:
```r
> x1 <- as_draws_df(example_draws())
> x2 <- as_draws_list(example_draws())

> x1_sub <- subset_draws(x1, chain = c(1, 1, 2), iteration = c(1:2, 1:50), unique = FALSE)
> x2_sub <- subset_draws(x2, chain = c(1, 1, 2), iteration = c(1:2, 1:50), unique = FALSE)

> rownames(x1_sub)
[1] "1"   "2"   "3"   "4"   "5"   "6"   "7"   "8"   "9"   "10"  ...
> rownames(as_draws_df(x2_sub))
[1] "1.1"  "1.2"  "1.3"  "1.4"  "1.5"  "1.6"  "1.7"  "1.8"  "1.9"  "1.10" ...
```
I'm nor sure which one is more desirable.

paul-buerkner:
Thanks! Since rownames don't tell us anything in this format, I think they should be as simple as possible. i.e. `NULL` or explicitely set to `1, 2, 3, ...` as in `x1_sub`.

tests/testthat/test-subset_draws.R
```diff
@@ -134,3 +134,14 @@ test_that("subset_draws preserves variable order", {
...
 
+test_that("non-unique subsetting for draws_df same as doing it with draws_list", {
+  x_df <- as_draws_df(example_draws())
+  x_list <- as_draws_list(example_draws())
+
+  x_df_sub <- subset_draws(x_df, chain = c(1,1,2), iteration = c(1:2, 1:50),
+                           unique = FALSE)
+  x_list_sub <- subset_draws(x_list, chain = c(1,1,2), iteration = c(1:2, 1:50),
+                             unique = FALSE)
+  expect_equivalent(x_df_sub, as_draws_df(x_list_sub))
```
jgabry, line 145: @paul-buerkner I had to use expect_equivalent here instead of expect_equal because the `as_draws_df(x_list_sub)` doesn't have row names. Do you know if that's intentional? It's only an issue after subsetting. Check this out: 

```r
x_df <- as_draws_df(example_draws())
x_list <- as_draws_list(example_draws())

# all good here
all.equal(rownames(x_df), rownames(as_draws_df(x_list))) 

x_df_sub <- subset_draws(x_df)          
x_list_sub <- subset_draws(x_list)

# now no row names for the second one
all.equal(rownames(x_df_sub), rownames(as_draws_df(x_list_sub)))
```

paul-buerkner: I don't think its intentional. in posterior, we don't really care about the rownames anywhere so I don't think it matter practically, but of course esthetically, a consistent behavior is desirble.

paul-buerkner: I am fine with sorting out the row names thing separately.

jgabry: Ok let's go ahead and merge this and I'll open a separate issue about the row names

jgabry:
Thanks again @Ozan147!

Issue #129, opened 2021-05-12
subset_draws can be prohibitively slow where regular subsetting isn't


jgabry:
If we have a lot of variables, e.g. 

```r
x <- matrix(rnorm(10 * 300000), nrow = 10, ncol = 300000)
x <- as_draws_matrix(x)
```

then regular R subsetting is super fast: 

```r
# super fast
x1 <- x[, colnames(x)]
```

but using `subset_draws()` is so slow I gave up and killed it before it finished:

```r
# broken
x2 <- subset_draws(x, variable = colnames(x))
```

paul-buerkner:
Sounds like another efficiency issue to tackle before release :-D

jgabry:
Yeah, hopefully we can track down the source of the inefficiency. I only realized this because of this issue with cmdstanr's csv reading (which calls subset_draws): 

https://discourse.mc-stan.org/t/cmdstanr-fails-to-read-its-own-csv-files-for-large-number-of-parameters/22450/4

jgabry:
I've been super busy lately with other stuff but I really need to make some time to work on posterior again! I'll dig into this as soon as I can make some time if it hasn't been solved by then.

paul-buerkner:
Thank you!

jgabry:
I ended up digging into this now and I figured out that the problem is coming from the internal `check_existing_variables` function:

```r
# this is just as slow as subset_draws() so I think this is where most of the issue is
x <- as_draws_matrix(matrix(rnorm(10 * 300000), nrow = 10, ncol = 300000))
check_existing_variables(colnames(x), x)
```

In particular it's this loop over the entire `variables` vector: 

https://github.com/stan-dev/posterior/blob/feeabf82514756fea7293d13d27aa59a1c537d44/R/draws-index.R#L442-L446

jgabry:
And also this loop, which only happens if `regex=TRUE`: 

https://github.com/stan-dev/posterior/blob/feeabf82514756fea7293d13d27aa59a1c537d44/R/draws-index.R#L430-L434

jgabry:
Unfortunately I tried switching to using `vapply` instead of looping but it doesn't speed it up nearly enough for it to be practical. 

I'm not sure what to do about this. If we can't speed up this check substantially we're going to need to turn it off or at least allow the user to turn it off because it's just not 300000 that's a problem, it's really slow even with only 5000 variables (not nearly as slow as with 300000 but still way to slow for just subsetting): 

```r
[...]

PR #130, opened 2021-05-13
make subset_draws efficient

Ozan147:
Addresses #129.

Makes `check_existing_variables()`, and therefore `subset_draws()` much more efficient. For example, with default arguments, this suggested version can process 300k variables in a fraction of a second while the current version takes almost 10 minutes. However, both versions still have comparable runtimes when  `regex = TRUE`.

```
x <- as_draws_matrix(matrix(rnorm(10 * 300000), nrow = 10, ncol = 300000))
microbenchmark::microbenchmark(
  posterior:::check_existing_variables(variables = colnames(x), x = x),
  posterior::subset_draws(x, variable = colnames(x)),
  times = 10,
  unit = "s"
)

Unit: seconds
       min        lq      mean    median        uq       max neval
 0.3240343 0.3354325 0.3669232 0.3541255 0.3870651 0.4765244    10
 0.4236527 0.4271074 0.4862439 0.4566513 0.5303358 0.6543742    10
```

### DIFF
--- a/R/draws-index.R
+++ b/R/draws-index.R
@@ -430,22 +430,21 @@ check_existing_variables <- function(variables, x, regex = FALSE,
   if (regex) {
     tmp <- named_list(variables)
     for (i in seq_along(variables)) {
-      tmp[[i]] <- all_variables[grepl(variables[i], all_variables)]
+      tmp[[i]] <- grep(variables[i], all_variables)
     }
     # regular expressions are not required to match anything
     missing_variables <- NULL
-    variables <- as.character(unique(unlist(tmp)))
+    variables <- as.character(all_variables[unique(unlist(tmp))])
   } else if (!scalar_only) {
-    tmp <- named_list(variables)
-    escaped_variables <- escape_all(variables)
-    missing <- rep(NA, length(variables))
-    for (i in seq_along(variables)) {
-      v_regex <- paste0("^", escaped_variables[i], "(\\[.*\\])?$")
-      tmp[[i]] <- all_variables[grepl(v_regex, all_variables)]
-      missing[i] <- !length(tmp[[i]])
-    }
-    missing_variables <- variables[missing]
-    variables <- as.character(unique(unlist(tmp)))
+    missing_candidates <- setdiff(variables, all_variables)
+    all_variables_base <- gsub("\\[.*\\]$", "", all_variables)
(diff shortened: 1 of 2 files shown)


rok-cesnovar:
Very nice!! Thanks so much!

jgabry:
Thanks so much @Ozan147! 

This looks good to me. I just added a test and now it's running the unit tests so I'll wait until those pass and will also let @paul-buerkner take a look at this before we merge. But I think this is fast enough now that we can keep the check (at least when regex=FALSE).

jgabry (approved):

paul-buerkner:
looks good to me. feel free to merge.

Jonah Gabry ***@***.***> schrieb am Do., 13. Mai 2021, 16:56:

mjskay:
You might be able to speed up the regex version by translating the input list of regexes into a single regex that matches if any one of them does.

paul-buerkner (approved):

jgabry:
Awesome, thanks @Ozan147! @paul-buerkner 

> You might be able to speed up the regex version by translating the input list of regexes into a single regex that matches if any one of them does.

good call! I'll try that. let's go ahead and merge so people can use the faster version on master and then I'll open an issue/PR just for fixing the regex version.

Ozan147:
@mjskay and @jgabry (#131), I tested the single regex pattern with `|` yesterday but it turns out there is a size limit. Even with `perl = TRUE`, it starts throwing errors after certain number of variables. One idea might be to divide that large pattern string into valid-size chunks to perform multiple matches.

mjskay:
Ah damn. Could also fall back to the for loop version in that case I suppose... Honestly I'd guess most use cases of the regex version have a small number of input regexes (else why use regex in the first place?)

jgabry:
Actually @Ozan147 this PR broke something that we need to fix (totally not your fault though because we didn't have a test for this). Before this PR the order of `variables` was preserved by `check_existing_variables`, but now it always sorts them: 

```r
# before 
> check_existing_variables(c("theta[2]", "theta[1]"), example_draws()
[1] "theta[2]" "theta[1]"

# now
> check_existing_variables(c("theta[2]", "theta[1]"), example_draws()
[1] "theta[1]" "theta[2]"
```

This causes some CmdStanR tests to fail.

jgabry:
I just added a test on master for this that's failing (which it should be given that it's not fixed yet)

PR #133, opened 2021-05-13
preserve order of variables in check_existing_variables

### LINKED ISSUE #132: subset_draws doesn't respect variable order anymore
Before PR #130 the order of `variables` was preserved by `check_existing_variables`, but now it always sorts them: 

```r
# before 
> check_existing_variables(c("theta[2]", "theta[1]"), example_draws())
[1] "theta[2]" "theta[1]"

# now
> check_existing_variables(c("theta[2]", "theta[1]"), example_draws())
[1] "theta[1]" "theta[2]"
```

This breaks the draws() method in CmdStanR and causes some CmdStanR tests to fail. It should have caused a test failure in posterior but we didn't have a test for this. I just added one that will fail until this is solved.

jgabry:
fixes #132

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: posterior
 Title: Tools for Working with Posterior Distributions
-Version: 0.1.4
-Date: 2021-03-26
+Version: 0.1.5
+Date: 2021-05-13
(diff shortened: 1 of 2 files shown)


jgabry:
Also bumped version number to 0.1.5 so we can differentiate between the versions before the speedup and after

rok-cesnovar (approved):

jgabry:
Going to merge this since it's passing CI and @rok-cesnovar reviewed it.

PR #136, opened 2021-05-14
updated drop_dims()

### LINKED ISSUE #135: `drop_dims()` doesn't work as expected
I've run into a couple of problems with `drop_dims()`:

- With default `dims = NULL`, it's throwing an error rather than dropping all single-level dimensions
- Its output breaks the `print.draws_array` method, which strictly expects a 3-dimensional `draws_array` input

```r
# select the first chain from the eight schools example
x <- example_draws()[, 1, ]
dim(x)
[1] 100   1  10

# default dims = NULL fails to drop dimensions with only one level
out <- posterior:::drop_dims(x, dims = NULL) 
# Error in dim(x) <- old_dims[-dims] : 
#   dims [product 10] do not match the length of object [1000]

# `print.draws_array` breaks when dims(draws_array) != 3
out <- posterior:::drop_dims(x, dims = 2) 
dim(out)
[1] 100  10
out
# Error in dimnames(x)[[3L]] : subscript out of bounds
```

Ozan147:
Fixes #135.

- With default dims = NULL, `drop_dims()` drops all single-level dimensions as expected
- `drop_dims()`'s output excludes the `draws_array` class and therefore doesn't use the `print.draws_array` method

### DIFF
--- a/R/rvar-slice.R
+++ b/R/rvar-slice.R
@@ -138,10 +138,9 @@
       if (all(names(new_dimnames) == "")) names(new_dimnames) <- NULL
       dimnames(x) <- new_dimnames
     }
-    x
-  } else {
-    x
   }
+
+  x
 }
 
 #' @export
--- a/R/rvar-slice.R
+++ b/R/rvar-slice.R
@@ -125,7 +125,20 @@
   ))
 
   if (drop) {
-    drop_(x)
+    .dim <- dim(x)
+    if (length(.dim) > 1) {
+      # with exactly 1 dimension left we don't want to drop anything
+      # (otherwise names get lost), so only do this with > 1 dimension
+      keep_dim <- .dim != 1
+      .dimnames <- dimnames(x)
+      dim(x) <- .dim[keep_dim]
+      # for comparison / testing, ensure if no dimnames have names that we
+      # actually have those names be NULL (rather than just empty strings)
+      new_dimnames <- .dimnames[keep_dim]
+      if (all(names(new_dimnames) == "")) names(new_dimnames) <- NULL
+      dimnames(x) <- new_dimnames
+    }
+    x
   } else {
     x
   }
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -8,7 +8,8 @@ Authors@R: c(person("Paul-Christian", "Bürkner", email = "paul.buerkner@gmail.c
              person("Aki", "Vehtari", email = "Aki.Vehtari@aalto.fi", role = c("aut")),
              person("Måns", "Magnusson", role = c("ctb")),
              person("Rok", "Češnovar", role = c("ctb")),
-             person("Ben", "Lambert", role = c("ctb")))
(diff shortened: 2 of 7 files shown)


paul-buerkner:
Thanks! I think we need to ensure all posterior classes are dropped if a dimension is dropped. That does mean we should handle the `draws_matrix` class as well, as well as remove the general `draws` class from the object.

Another questions is, if we still want to drop classes even if no dimension was dropped? I can think of arguments for both ways.

Ozan147:
Modified the drop function to remove dimensions selectively **and/or** get rid of posterior classes. After some exploration, using an optional boolean parameter to set classes to NULL and let R decide appropriate native classes worked best for me. This also allows users to reset classes even if no dimension was dropped.

Since the function is now doing more than just dropping dimensions, I renamed it from `drop_dims()` to `drop2()` following your conventions with other functions (`stop2()`, `deparse2()`, etc.). It mirrors `base::drop()` by default but provides more flexibility with additional parameters. 

There was a separate `drop_()` helper within `rvar-slice` tools. I first tried to merge both functions but it wasn't clean due to `rvar` objects not being of `array` class and having some edge cases. Then I realized `drop_()` was a small function that was getting called only once, so I got rid of it to prevent confusion (assuming it is not used in any downstream packages as `rvar` already has a slicer method with drop option).

Let me know what you think about these changes. I can keep iterating until we are happy with it.

R/rvar-slice.R
```diff
@@ -125,7 +125,20 @@
...
+      # (otherwise names get lost), so only do this with > 1 dimension
+      keep_dim <- .dim != 1
+      .dimnames <- dimnames(x)
+      dim(x) <- .dim[keep_dim]
+      # for comparison / testing, ensure if no dimnames have names that we
+      # actually have those names be NULL (rather than just empty strings)
+      new_dimnames <- .dimnames[keep_dim]
+      if (all(names(new_dimnames) == "")) names(new_dimnames) <- NULL
+      dimnames(x) <- new_dimnames
+    }
```
mjskay, line 140: > There was a separate drop_() helper within rvar-slice tools. I first tried to merge both functions but it wasn't clean due to rvar objects not being of array class and having some edge cases. Then I realized drop_() was a small function that was getting called only once, so I got rid of it to prevent confusion (assuming it is not used in any downstream packages as rvar already has a slicer method with drop option).

That's fine with me. I had thought I would need that function in more than one spot in rvar internals and then I didn't. Stylistically I would probably change this:

```r
    x
  } else {
    x
  }
```

to this:

```r
  }

  x
```

Ozan147: Updated it, thanks! Let me know if there is any other change you'd like.

mjskay: Thanks!

mjskay:
Changes on the rvar end of things look good to me, thanks!

paul-buerkner:
Thanks! Looks good to me too. Merging now.

PR #137, opened 2021-05-16
use more informative internal function names

jgabry:
I think it will make our source code easier to read if function names indicate their purpose as much as possible. Most of our function names already do this well, but this PR changes a few that don't: 

`deparse2` -> `deparse_pretty` 
`drop2` -> `drop_dims_or_classes`
`stop2` -> `stop_no_call` 

I'm not attached to these particular new names so I'm happy to change them if anyone has suggestions, but I think these names at least hint at the reason for the existence of the functions.

### DIFF
--- a/R/as_draws.R
+++ b/R/as_draws.R
@@ -65,7 +65,7 @@ closest_draws_format <- function(x) {
     out <- "list"
   }
   else {
-    stop2("Don't know how to transform an object of class ",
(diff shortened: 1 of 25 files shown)


paul-buerkner:
Looks good to me. Feel free to merge.

jgabry:
Ok thanks will go ahead and merge. If anyone else has ideas for better names than these we can open another PR and change them.

PR #138, opened 2021-05-17
Improve summarise_draws efficiency

jsocolar:
This can close https://github.com/stan-dev/posterior/issues/98.

`summarise_draws` now internally writes output to a preallocated matrix, which makes the "chunking" option unnecessary except for parallelization, which will be dealt with separately.  

Additionally, we now check whether
```
  if ("variable" %in% the_names) {
    stop2("Name 'variable' is reserved in 'summarise_draws'.")
  }
```
prior to the potentially lengthy computation.

Benchmarks are:

```
remotes::install_github("jsocolar/posterior")
library(posterior)
library(microbenchmark)

set.seed(1)
nc <- 4

n <- 100
test_array <- array(data = rnorm(1000*nc*n), dim = c(1000,nc,n))
y <- posterior::as_draws_array(test_array)

b100 <- microbenchmark(m_s <- posterior::summarise_draws(y), times = 10)
# .786 sec under current posterior
# .79 sec under new version

n <- 1000
test_array <- array(data = rnorm(1000*nc*n), dim = c(1000,nc,n))
y <- posterior::as_draws_array(test_array)

b1000 <- microbenchmark(m_s <- posterior::summarise_draws(y), times = 2)
# 7.5 sec under current posterior
# 7.5 sec under new version

n <- 10000
test_array <- array(data = rnorm(1000*nc*n), dim = c(1000,nc,n))
y <- posterior::as_draws_array(test_array)

b10000 <- microbenchmark(m_s <- posterior::summarise_draws(y), times = 1)
# 84 under current posterior
# 73.5 sec under new version

n <- 50000
test_array <- array(data = rnorm(1000*nc*n), dim = c(1000,nc,n))
y <- posterior::as_draws_array(test_array)

b50000 <- microbenchmark(m_s <- posterior::summarise_draws(y), times = 1)
# 701 sec under current posterior
# 385 sec under new version
```

### DIFF
--- a/R/summarise_draws.R
+++ b/R/summarise_draws.R
@@ -127,7 +127,7 @@ summarise_draws.draws <- function(x, ..., .args = list()) {
 
   # get length and output names, calculated on the first variable
   v1 <- variables[1]
-  draws1 <- drop_dims_or_classes(x[, , v1], dims = 3)
+  draws1 <- drop_dims_or_classes(x[, , v1], dims = 3, reset_class = FALSE)
   args1 <- c(list(draws1), .args)
   out1 <- named_list(names(funs))
   n_fun <- length(out1)
--- a/R/summarise_draws.R
+++ b/R/summarise_draws.R
@@ -133,16 +133,17 @@ summarise_draws.draws <- function(x, ..., .args = list()) {
   for (m in names(funs)) {
     out1[[m]] <- do.call(funs[[m]], args1)
   }
-  the_names <- vector()
+  the_names <- vector(mode = "list", length = length(out1))
   for (i in 1:length(out1)){
     if (rlang::is_named(out1[[i]])) {
-      the_names <- c(the_names, names(out1[[i]]))
+      the_names[[i]] <- names(out1[[i]])
     } else if (length(out1[[i]]) > 1) {
-      the_names <- c(the_names, paste0(names(out1)[i], ".", c(1:length(out1[[i]]))))
+      the_names[[i]] <- paste0(names(out1)[i], ".", c(1:length(out1[[i]])))
     } else {
-      the_names <- c(the_names, names(out1)[i])
+      the_names[[i]] <- names(out1)[i]
     }
   }
+  the_names <- unlist(the_names)
   
   # Check for naming issues prior do doing lengthy computation
   if ("variable" %in% the_names) {
--- a/R/summarise_draws.R
+++ b/R/summarise_draws.R
@@ -68,7 +68,6 @@ summarise_draws.default <- function(x, ...) {
 #' @rdname draws_summary
 #' @export
 summarise_draws.draws <- function(x, ..., .args = list()) {
-  print("new version")
   funs <- as.list(c(...))
   .args <- as.list(.args)
   if (length(funs)) {
--- a/R/summarise_draws.R
+++ b/R/summarise_draws.R
@@ -124,23 +124,50 @@ summarise_draws.draws <- function(x, ..., .args = list()) {
   x <- repair_draws(x)
   x <- as_draws_array(x)
   variables <- variables(x)
-  out <- named_list(variables, values = list(named_list(names(funs))))
-  for (v in variables) {
-    draws <- drop_dims(x[, , v], dims = 3)
-    args <- c(list(draws), .args)
-    for (m in names(funs)) {
-      out[[v]][[m]] <- do.call(funs[[m]], args)
-      if (rlang::is_named(out[[v]][[m]])) {
-        # use returned names to label columns
-        out[[v]][[m]] <- rbind(out[[v]][[m]])
-      }
-    }
-    out[[v]] <- do.call(cbind, out[[v]])
+  # get length and output names, calculated on the first variable
+  v1 <- variables[1]
+  draws1 <- drop_dims(x[, , v], dims = 3)
+  args1 <- c(list(draws1), .args)
+  out1 <- named_list(names(funs))
+  for (m in names(funs)) {
+    out1[[m]] <- do.call(funs[[m]], args1)
   }
-  out <- tibble::as_tibble(do.call(rbind, out))
-  if ("variable" %in% names(out)) {
+  the_names <- vector()
+  for (i in 1:length(out1)){
+    if (rlang::is_named(out1[[i]])) {
+      the_names <- c(the_names, names(out1[[i]]))
+    } else if (length(out1[[i]]) > 1) {
+      the_names <- c(the_names, paste0(names(out1)[i], ".", c(1:length(out1[[i]]))))
+    } else {
+      the_names <- c(the_names, names(out1)[i])
+    }
+  }
+  
+  # Check for naming issues prior do doing lengthy computation
+  if ("variable" %in% the_names) {
     stop2("Name 'variable' is reserved in 'summarise_draws'.")
   }
+  
+  # Pre-allocate matrix to store output
+  out <- matrix(NA, nrow = length(variables), ncol = length(unlist(out1)))
+  colnames(out) <- the_names
+  out[1,] <- unlist(out1)
+  
+  # Do the computation for all remaining variables
+  if(length(variables) > 1L){
+    for (v_ind in 2:length(variables)) {
+      v <- variables[v_ind]
+      draws <- drop_dims(x[, , v], dims = 3)
+      args <- c(list(draws), .args)
+      out_v <- named_list(names(funs))
+      for (m in names(funs)) {
+        out_v[[m]] <- do.call(funs[[m]], args1)
+      }
+      out[v_ind, ] <- unlist(out_v)
+    }
+  }
+  
+  out <- tibble::as_tibble(out)
   out$variable <- variables
   out <- move_to_start(out, "variable")
   class(out) <- class_draws_summary()
--- a/R/summarise_draws.R
+++ b/R/summarise_draws.R
@@ -152,7 +152,7 @@ summarise_draws.draws <- function(x, ..., .args = list()) {
   }
   
   # Pre-allocate matrix to store output
-  out <- matrix(NA, nrow = length(variables), ncol = length(unlist(out1)))
+  out <- matrix(NA, nrow = length(variables), ncol = length(the_names))
   colnames(out) <- the_names
   out[1, ] <- unlist(out1)
   
--- a/R/summarise_draws.R
+++ b/R/summarise_draws.R
@@ -130,11 +130,12 @@ summarise_draws.draws <- function(x, ..., .args = list()) {
   draws1 <- drop_dims_or_classes(x[, , v1], dims = 3)
   args1 <- c(list(draws1), .args)
   out1 <- named_list(names(funs))
+  n_fun <- length(out1)
   for (m in names(funs)) {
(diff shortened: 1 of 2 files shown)


R/summarise_draws.R
```diff
@@ -124,23 +124,51 @@ summarise_draws.draws <- function(x, ..., .args = list()) {
...
+
+  # get length and output names, calculated on the first variable
+  v1 <- variables[1]
+  draws1 <- drop_dims_or_classes(x[, , v1], dims = 3)
+  args1 <- c(list(draws1), .args)
+  out1 <- named_list(names(funs))
+  for (m in names(funs)) {
+    out1[[m]] <- do.call(funs[[m]], args1)
+  }
+  the_names <- vector()
```
paul-buerkner, line 136: it might be more efficient to first gather names in a list of length `out1` and then unlist at the end instead of appending to `the_names` starting from a zero-length vector.

R/summarise_draws.R
```diff
@@ -124,23 +124,51 @@ summarise_draws.draws <- function(x, ..., .args = list()) {
...
-  out <- tibble::as_tibble(do.call(rbind, out))
-  if ("variable" %in% names(out)) {
+  
+  # Check for naming issues prior do doing lengthy computation
+  if ("variable" %in% the_names) {
     stop_no_call("Name 'variable' is reserved in 'summarise_draws'.")
   }
+  
+  # Pre-allocate matrix to store output
+  out <- matrix(NA, nrow = length(variables), ncol = length(unlist(out1)))
```
paul-buerkner, line 153: right now, `out1` is unlisted multiple times here. Perhaps do it only once? For example, `ncol` could directly be `length(the_names)`.

R/summarise_draws.R
```diff
@@ -124,23 +124,51 @@ summarise_draws.draws <- function(x, ..., .args = list()) {
...
+  # get length and output names, calculated on the first variable
+  v1 <- variables[1]
+  draws1 <- drop_dims_or_classes(x[, , v1], dims = 3)
+  args1 <- c(list(draws1), .args)
+  out1 <- named_list(names(funs))
+  for (m in names(funs)) {
+    out1[[m]] <- do.call(funs[[m]], args1)
+  }
+  the_names <- vector()
+  for (i in 1:length(out1)){
```
paul-buerkner, line 137: shall we use `seq_along(out1)` instead, to also handle the `length(out1) == 0L` case consistently?

R/summarise_draws.R
```diff
@@ -124,23 +124,51 @@ summarise_draws.draws <- function(x, ..., .args = list()) {
...
-    args <- c(list(draws), .args)
-    for (m in names(funs)) {
-      out[[v]][[m]] <- do.call(funs[[m]], args)
-      if (rlang::is_named(out[[v]][[m]])) {
-        # use returned names to label columns
-        out[[v]][[m]] <- rbind(out[[v]][[m]])
-      }
+
+  # get length and output names, calculated on the first variable
+  v1 <- variables[1]
```
paul-buerkner, line 129: do we know at this time that we have at least one variable in the draws object?

R/summarise_draws.R
```diff
@@ -124,23 +124,51 @@ summarise_draws.draws <- function(x, ..., .args = list()) {
...
+  draws1 <- drop_dims_or_classes(x[, , v1], dims = 3)
```
paul-buerkner, line 130: I suggest adding `reset_class = FALSE` as in the removed code.

R/summarise_draws.R
```diff
@@ -124,23 +124,51 @@ summarise_draws.draws <- function(x, ..., .args = list()) {
...
+  # Pre-allocate matrix to store output
+  out <- matrix(NA, nrow = length(variables), ncol = length(unlist(out1)))
+  colnames(out) <- the_names
+  out[1,] <- unlist(out1)
+  
+  # Do the computation for all remaining variables
+  if(length(variables) > 1L){
+    for (v_ind in 2:length(variables)) {
+      v <- variables[v_ind]
+      draws <- drop_dims_or_classes(x[, , v], dims = 3)
```
paul-buerkner, line 161: since we now use these few number of lines twice, once for the first variable and the second time for the rest, is it worth making a small helper function out of it?

R/summarise_draws.R
```diff
@@ -124,23 +124,51 @@ summarise_draws.draws <- function(x, ..., .args = list()) {
...
     stop_no_call("Name 'variable' is reserved in 'summarise_draws'.")
   }
+  
+  # Pre-allocate matrix to store output
+  out <- matrix(NA, nrow = length(variables), ncol = length(unlist(out1)))
+  colnames(out) <- the_names
+  out[1,] <- unlist(out1)
+  
+  # Do the computation for all remaining variables
+  if(length(variables) > 1L){
```
paul-buerkner, line 158: add a space after `if` and before `{` for consistency with the existing code base.

R/summarise_draws.R
```diff
@@ -124,23 +124,51 @@ summarise_draws.draws <- function(x, ..., .args = list()) {
...
+  
+  # Check for naming issues prior do doing lengthy computation
+  if ("variable" %in% the_names) {
     stop_no_call("Name 'variable' is reserved in 'summarise_draws'.")
   }
+  
+  # Pre-allocate matrix to store output
+  out <- matrix(NA, nrow = length(variables), ncol = length(unlist(out1)))
+  colnames(out) <- the_names
+  out[1,] <- unlist(out1)
```
paul-buerkner, line 155: add a space after `,` for consistency with the existing code base.

paul-buerkner (changes requested):
Thank you for working on this! The speed improvements are already quite noticable when using a lot of variables. I have a few minor change requests but only edge case stuff and minor stylistic things.

R/summarise_draws.R
jsocolar: Would you prefer that the helper function be defined inside `summarise_draws.draws`, in summarise_draws.R but outside of any existing functions, or somewhere else in the codebase?

paul-buerkner: Both, inside summarise_draws.draws, in summarise_draws.R would be fine for me.

R/summarise_draws.R
jsocolar: We don't. Current behavior returns a 0x1 tibble in this case.  Shall I just hard-code this output?  

```
nc <- 4
n <- 1
test_array <- array(data = rnorm(1000*nc*n), dim = c(1000,nc,n))
y <- posterior::as_draws_array(test_array)
dimnames(y)$variable[1] <- ".log_weight"
posterior::variables(y)
posterior::summarise_draws(y)
```

R/summarise_draws.R
jsocolar: Yeah good catch.  This was just me rushing too fast while resolving conflicts.

R/summarise_draws.R
paul-buerkner: I think it would be desireble to hard-code it, yes, or to throw an error even? Not sure how a consistent output would look like even.

jsocolar: I've hard-coded with a warning just in case the current behavior is actually necessary for somebody inside a giant for-loop where adding the necessary `tryCatch` would be a pain

jsocolar:
@paul-buerkner This should be ready for another look.

jgabry:
Thanks @jsocolar! will take a look shortly.

jgabry (changes requested):
This is great @jsocolar! Thanks a ton for working on this. I think the implementation looks correct and all of my suggestions in the review comments are small and really they're super nitpicky ;)

jgabry:
Also it looks like the tests are passing, so that's great. (You can ignore the failure on R-CMD-check / macOS-latest (devel) because there's some issue with R-devel on Mac right now that has nothing to do with this package)

R/summarise_draws.R
```diff
@@ -124,27 +124,55 @@ summarise_draws.draws <- function(x, ..., .args = list()) {
...
+      }
+    }
+    out <- tibble::as_tibble(out)
+    out$variable <- variables
+    out <- move_to_start(out, "variable")
+    class(out) <- class_draws_summary()
+    return(out)
+  } else { # i.e. if length(variables) == 0L
+    warning("the draws object contained no variables with unreserved names; 
+            no summaries computed")
```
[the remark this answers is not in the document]
jgabry: I think we should also add a test to `test-summarise-draws.R` to make this warning is thrown when it should be

R/summarise_draws.R
```diff
@@ -124,27 +124,55 @@ summarise_draws.draws <- function(x, ..., .args = list()) {
...
+    for (i in seq_along(out1)){
+      if (rlang::is_named(out1[[i]])) {
+        the_names[[i]] <- names(out1[[i]])
+      } else if (length(out1[[i]]) > 1) {
+        the_names[[i]] <- paste0(names(out1)[i], ".", c(1:length(out1[[i]])))
+      } else {
+        the_names[[i]] <- names(out1)[i]
+      }
+    }
+    the_names <- unlist(the_names)
```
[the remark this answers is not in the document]
jsocolar: I don't think so.  The problem is if the length of the function output is > 1 (as for `quantile2`).  Originally I initialized a character vector of length 0 and grew it iteratively through the loop with a bunch of `c()` calls, but Paul preferred to preallocate, which is where this came from.
Or am I being dense here?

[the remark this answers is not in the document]
jgabry: No I think you're right because of the output > 1 case. I forgot about that.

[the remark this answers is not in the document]
jgabry: Yeah I double checked and you're definitely right about this. Forget about my suggestion here

jgabry (approved):
Thanks for addressing the comments so quickly. This looks good to me!

paul-buerkner (approved):

paul-buerkner:
This is also good from my part. Feel free to merge.

jgabry:
I'll quickly add a test for the new warning message and then I'll merge it

Ozan147:
I was wondering if you have a merge strategy like squash to condense commits in master 🤔

jgabry:
> I was wondering if you have a merge strategy like squash to condense commits in master 🤔

We haven't been doing that but perhaps we should

jgabry:
> I was wondering if you have a merge strategy like squash to condense commits in master 🤔

On the one hand condensing is nice, on the other hand it's nice to be able to see the full history (i.e., all individual commits). But I'm open to changing what we've been doing. If you want you can open an issue about this and we can decide on the strategy we want to use going forward.

Ozan147:
Yeah, there is certainly a trade-off there. Even without setting an official blanket strategy in the repo, individual contributors could selectively squash their trivial commits to keep things a little cleaner. I can create an issue to discuss and maybe create guidelines but obviously it's nothing pressing 👍

jgabry:
> Yeah, there is certainly a trade-off there. Even without setting an official blanket strategy in the repo, individual contributors could selectively squash their trivial commits to keep things a little cleaner. I can create an issue to discuss and maybe create guidelines but obviously it's nothing pressing 👍

Awesome, thanks!

jgabry:
Will merge now since it's passing CI (except for R devel on Mac which isn't related to this). Thanks again @jsocolar!

PR #139, opened 2021-05-17
Updated as_draws_df()

### LINKED ISSUE #134: after subsetting row names don't match when converting list to df
```r
x_df <- as_draws_df(example_draws())
x_list <- as_draws_list(example_draws())

# all good here
all.equal(rownames(x_df), rownames(as_draws_df(x_list))) 

# now call subset_draws
x_df_sub <- subset_draws(x_df)          
x_list_sub <- subset_draws(x_list)

# "400 string mismatches"
all.equal(rownames(x_df_sub), rownames(as_draws_df(x_list_sub)))
```

This doesn't matter a ton since we don't use the row names for anything meaningful, but I guess the question is whether we should make the row names of `as_draws_df(x_list_sub)` the same as the ones of `x_df_sub` or if we should NULL them out in both cases.

Ozan147:
Fixes #134 and and speeds up `as_draws_df()`:
- makes the `as_draws_df` methods vectorized, faster and more concise
- resolves row name inconsistencies that occur after transformations from various classes to `draws_df`
- switches to invisible returns for `check_*_*()` functions to prevent their output from getting printed
- enforces equal input lengths in `compute_draw_ids()` to avoid incorrect outputs due to recycling of vector elements
- applies more strict tests in relevant parts

### DIFF
--- a/R/as_draws_df.R
+++ b/R/as_draws_df.R
@@ -61,13 +61,12 @@ as_draws_df.draws_df <- function(x, ...) {
 #' @rdname draws_df
 #' @export
 as_draws_df.draws_matrix <- function(x, ...) {
-  class(x) <- "matrix"
-  draws <- as.integer(rownames(x))
-  rownames(x) <- NULL
+  if (ndraws(x) == 0L) {
+    return(empty_draws_df(variables(x)))
+  }
   x <- tibble::as_tibble(x)
-  x$.chain <- rep(1L, nrow(x))
-  x$.iteration <- draws
-  x$.draw <- draws
+  x[".chain"] <- 1L
+  x[c(".iteration", ".draw")] <- 1L:nrow(x)
   class(x) <- class_draws_df()
   x
 }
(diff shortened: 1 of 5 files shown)


R/as_draws_df.R
```diff
@@ -61,58 +61,50 @@ as_draws_df.draws_df <- function(x, ...) {
...
-    out[[i]]$.iteration <- iteration_ids
-    out[[i]]$.draw <- compute_draw_ids(chain_ids[i], iteration_ids)
-  }
-  out <- do.call(rbind, out)
-  class(out) <- class_draws_df()
-  out
+  x <- apply(x, 3L, c)
+  x <- tibble::as_tibble(x)
+  x[".chain"] <- rep(chain_ids, each = max(iteration_ids))
+  x[".iteration"] <- rep(iteration_ids, max(chain_ids))
```
Ozan147, line 85: This vectorization might cause a problem only if the `draws_array` input is skipping a draw or has randomly ordered draws but I was not able to produce that specific case. In my attempts to create edge cases to break this function, chain/iteration/draw ids are getting repaired or I get errors indicating that such `draws_array` structures are not possible.

R/as_draws_df.R
```diff
@@ -61,58 +61,50 @@ as_draws_df.draws_df <- function(x, ...) {
...
-  }
-  out <- do.call(rbind, out)
-  class(out) <- class_draws_df()
-  out
+  vars <- names(x[[1L]])
+  x <- do.call(rbind.data.frame, x)
+  colnames(x) <- vars
+  x <- tibble::as_tibble(x)
+  x[".chain"] <- rep(chain_ids, each = max(iteration_ids))
+  x[".iteration"] <- rep(iteration_ids, max(chain_ids))
```
Ozan147, line 104: Similarly, this vectorization might cause a problem only if the `draws_list` input is skipping a draw or has randomly ordered draws but I was not able to produce that specific case. In my attempts to create edge cases to break this function, chain/iteration/draw ids are getting repaired or I get errors indicating that such `draws_list` structures are not possible.

R/repair_draws.R
```diff
@@ -140,6 +140,7 @@ repair_chain_ids <- function(chain_ids) {
 #' @param iteration_ids A vector of iteration indices
 #' @noRd
 compute_draw_ids <- function(chain_ids, iteration_ids) {
+  assert_true(length(chain_ids) == length(iteration_ids))
```
Ozan147, line 143: Without this assertion, there is room for error:
```r
compute_draw_ids(chain_ids = 1:2, iteration_ids = 1:10)
 [1]  1 12  3 14  5 16  7 18  9 20
```

Ozan147 (commented):
Although, these updates seem to pass the official tests in the repo as well as my local tests, feel free to try them with any additional edge cases to make sure everything works as expected.

R/as_draws_df.R
paul-buerkner: It looks fine for me too.

R/as_draws_df.R
paul-buerkner: Also looks fine for me.

R/repair_draws.R
paul-buerkner: Thanks, Good catch!

R/as_draws_df.R
```diff
@@ -154,9 +146,9 @@ as_draws_df.mcmc.list <- function(x, ...) {
   }
   if (has_iteration_column) {
     iteration_ids <- x[[.iteration]]
-    x[[.iteration]] <- NULL
+    x[.iteration] <- NULL
   } else {
-    iteration_ids <- seq_len(NROW(x))
+    iteration_ids <- 1L:nrow(x)
```
paul-buerkner, line 151: What is the reason for this change? `seq_len(n)` is usually preferred over `1L:n` because the former handles `n = 0` consistently.

R/as_draws_df.R
```diff
@@ -171,22 +163,22 @@ as_draws_df.mcmc.list <- function(x, ...) {
...
   }
   # prepare draw indices --- i.e. drop them, since they are regenerated below
-  x[[".draw"]] <- NULL
+  x[".draw"] <- NULL
 
   # add reserved variables to the data
   check_new_variables(names(x))
-  x$.chain <- chain_ids
-  x$.iteration <- iteration_ids
+  x[".chain"] <- chain_ids
```
paul-buerkner, line 175: Why these changes from `$` to `[ ]`? As far as I can tell, these are always equivalent in the assignment case and also equivalent in the extraction case if we know the columns to exist in the df?

R/as_draws_df.R
Ozan147: I thought `1L:n` would be slightly faster but `seq_len(n)` is more robust and better choice.

R/as_draws_df.R
Ozan147: Yeah, they are equivalent. For some reason, I always thought using brackets + name string was better practice in package/shiny development but cannot find a reference at the moment ([other than this old article](https://www.r-bloggers.com/2009/10/r-accessors-explained/)) to support it.

Would you like me to revert these?

paul-buerkner: Thanks! It is fine for me to leave it like this, just wanted to understand if there was a reason I was not aware of.

jgabry: > I always thought using brackets + name string was better practice in package/shiny development

I think you're right because using `$` allows for partial matching, which is a bit dangerous. In this case I don't think there's a chance of partial matching so it doesn't matter, but in general I think this is why `[` is preferred to `$`.

jgabry:
Thanks Ozan! I was going to suggest adding you as an official contributor but looks like @paul-buerkner already added you to the DESCRIPTION file after your last PR (thanks Paul). Thanks for all these contributions. 

R-Cmd-check on macOS-latest (devel) is failing but it's not related to this PR (also happening in other repos because of a failure to install R-devel).

paul-buerkner:
the PR looks ready for me. @jgabry feel free to merge if you agree.

jgabry (approved):

jgabry:
I think it's good to go. Will merge now since it's passing everything except for Mac r-devel for unrelated reasons.

Issue #124, opened 2021-04-18
Adding multi-core functionality to summarize_draws


saudiwin:
Hi - 

This certainly falls into the realm of feature request. I am using posterior and enjoying the functionality, but when there are a lot of parameters some functions like `summarize_draws` (not apologizing for the American spellling :) can take a long time. Given that these are doing parameter by parameter calculations, would it be possible to convert one of the loops to be an mclapply loop with an optional mc.cores argument? Could save a lot of time with models with lots of parameters.

Also here is some code that implements this. If there is interest I can do a PR:

```
library(posterior)
library(dplyr)

obj <- example_draws()

summarize_draws_mc <- function(obj,  ..., cores=parallel::detectCores()) {
  
  varlist <- dimnames(obj)$variable
  
  # find an optimal-ish breakdown of variable names
  
  var_break <- floor(length(varlist)/cores)
  var_extra <- length(varlist) %% cores
  
  var_splits <- c(rep(1:cores,each=var_break),
                  1:var_extra)
  
  
  over_vars <- parallel::mclapply(1:cores, function(c,...){
    
    subset_draws(obj,variable=varlist[var_splits==c]) %>% 
      summarize_draws(...)
    
  },mc.cores=cores) %>% 
    bind_rows
  
  slice(over_vars,match(varlist,variable))
  
}

sum_obj <- summarize_draws(obj)
sum_obj_mc <- summarize_draws_mc(obj)

identical(sum_obj,sum_obj_mc)

#check time differences

time1 <- system.time(summarize_draws(obj))
time2 <- system.time(summarize_draws_mc(obj))

time1
time2

# not particularly efficient, but should speed up with more variables/parameters

```

jgabry:
Yeah I think parallelization is something we should consider. Anyone have a reason why we shouldn't do this?

jsocolar:
This is closely related to https://github.com/stan-dev/posterior/issues/98.  It turns out that there's a big speedup available even from splitting up the computation *without* using multiple cores.  I'll have a bit more on that issue shortly.

saudiwin:
I think that @jsocolar re-writing of the function would make it much more scalable. I do still think that multicore functionality would still be useful given how easy it is to implement esp. with the R parallel package. It is true it won't work on Windows but it's also super easy to code. Of course I can always maintain a separate wrapper even if it's not in the main package.

jsocolar:
[...]

PR #141, opened 2021-05-18
Parallelized version of summarise_draws()

jsocolar:
This passes tests for me locally on OSX, and I have verified equivalency of output to the old `summarise_draws`.

Can close https://github.com/stan-dev/posterior/issues/124


jsocolar:
Checks are failing apparently because multicore is a no-go.  I can remove the multi-core tests if necessary?  Or maybe it's just a matter of adding `parallel` to imports?

paul-buerkner:
not sure but I would in any case recommend to at least not run multi core
tests on cran. locally they should run fine I think(?)

Jacob B. Socolar ***@***.***> schrieb am Di., 18. Mai 2021,
17:19:

rok-cesnovar:
@jsocolar do they run locally? Github Actions instances are typically single core I think, maybe that is why?

jsocolar:
Yeah they're fine locally.  Not sure how to specify not to run on github.

jsocolar:
Could just comment them out?

jsocolar:
Somebody should also check locally on Windows and Linux (not an option for me).

rok-cesnovar:
Based on this: https://stackoverflow.com/a/45294642 we could try setting `options(mc.cores=2)`

jsocolar:
I'll try dropping cores from 4 to 2 in the test suite.

jsocolar:
This seems to work, but now I'm failing a check for a real reason, so I'll do some digging.

tests/testthat/test-summarise_draws.R
```diff
@@ -68,3 +68,32 @@ test_that("summarise_draws warns if all variable names are reserved", {
...
+            x <- as_draws_array(test_array)
+            sum_x <- summarise_draws(x)
+            parsum_x <- summarise_draws(x, cores = 2)
+            expect_identical(sum_x, parsum_x)
+            
+            dimnames(x)$variable[1] <- reserved_variables()[1]
+            suppressWarnings(sum_x <- summarise_draws(x))
+            suppressWarnings(parsum_x <- summarise_draws(x, cores = 2))
+            expect_identical(sum_x, parsum_x)
+})
```
jgabry, line 99: These tests look good. I think it would also be good to test that it errors for invalid `cores` specifications.

tests/testthat/test-summarise_draws.R
```diff
@@ -68,3 +68,32 @@ test_that("summarise_draws warns if all variable names are reserved", {
   variables(x) <- ".log_weight"
   expect_warning(summarize_draws(x), "no variables with unreserved names")
 })
+
+test_that("multicore summarise_draws is identical to single-core summarise_draws 
+          including if some chunks contain no variables", {
+            set.seed(1)
+            nc <- 4
+            n <- 2
```
jgabry, line 76: I think some tests with `n > 2` would help make sure that the chunking works properly, right?

jgabry (commented):
I haven't actually tried this out yet but here are some preliminary comments on the code (which looks good). I'll try to review it more thoroughly soon. Hopefully others can review it too.

jgabry:
> Yeah they're fine locally. Not sure how to specify not to run on github.

For future reference, you can use `testthat::skip_on_ci()` to skip a given test on github actions. But I think you don't need that here because 2 cores should be ok.

jgabry:
Looks like it's super annoying to get working on Windows, sorry about that! Let us know if you want any help with that. I'm pretty useless with Windows (although I can certainly try to help) but I know both @paul-buerkner and @rok-cesnovar use Windows.

jsocolar:
Much of the slog is just because I'm waiting for the checks to run on github after ever bit of fiddling, since it's my only way to run the tests on windows :P.  And that's taking for ever (like 20 minutes each time), I suspect because github's in no hurry to give me extra cores to play with.  Annoyingly, it's just the windows checks that are slow.

jgabry:
I think part of it is that caching of dependencies doesn't work on windows so they're re-installed every time.

jgabry:
If you want you can just hold off on fixing the windows issue and someone else who can run windows tests locally can try to fix it.

jsocolar:
Ha, yeah, i might give one or two more attempts.  I've got other stuff to do while this runs in the background.  The challenge is getting internal functions from posterior passed out to the PSOCK cluster in a way that R-CMD-check is ok with.

jsocolar:
Still working on getting the PSOCK cluster working properly for windows.  Once that's working I'll implement @jgabry's outstanding suggestions. I want to flag a couple of residual issues:
1. I’ve added documentation of the cores argument in `summarise_draws.R` (see line 21), but I cannot figure out how to get this to show up in the docs at `draws_summary.html`.  Pointers appreciated.
1. To import `parallel`, would I just add `#' @import parallel` somewhere in `summarise_draws.R` or is there a different file where you keep all the dependencies stuff?
1. the GitHub checks on windows now take a long time (c. 20 minutes) to complete.  I think that’s because we are waiting for a long time for the extra cores to become available.  This is kind of a drag. Once we know they pass, would you like me to disable those tests on GitHub, or leave them be?
1. Should I add an option to use a socket cluster on Mac/Linux for users who don’t want to fork from within Rstudio?

jgabry:
> 1. I’ve added documentation of the cores argument in `summarise_draws.R` (see line 21), but I cannot figure out how to get this to show up in the docs at `draws_summary.html`.  Pointers appreciated.

Will take a look shortly. 

> 2. To import `parallel`, would I just add `#' @import parallel` somewhere in `summarise_draws.R` or is there a different file where you keep all the dependencies stuff?

That would import all of parallel's exported functions into our namespace. Since we're just using a couple of functions all you need to do is: 

* Add parallel to the Imports field in the DESCRIPTION file. Or, if we're only using it if `cores > 1`, add it to the Suggests field. If you put it in Suggests in theory you should then add a check for it being installed before it's used but in this case I think everyone who has R automatically has parallel. 
* Use `parallel::foo` to call `foo` 

> 3. the GitHub checks on windows now take a long time (c. 20 minutes) to complete.  I think that’s because we are waiting for a long time for the extra cores to become available.  This is kind of a drag. Once we know they pass, would you like me to disable the multicore tests on GitHub, or leave them be?

I'm not sure it's the multiple cores, I think it's mostly due to Windows not caching dependencies.  In [macOS-latest (release)](https://github.com/stan-dev/posterior/pull/141/checks?check_run_id=2614856613) the Install Dependencies step took only 5s (because they're cached). In [windows-latest (release)](https://github.com/stan-dev/posterior/pull/141/checks?check_run_id=2614856639) the Install Dependencies step took 18m 14s. It seems to be compiling a bunch of stuff for some reason (but I would have expected it to use Windows binaries), but I just glanced at it briefly so not sure why it's taking so long. I guess you  can test it by turning off the multiple cores and seeing if that speeds it up. But I suspect it won't unless that somehow shortens the Install Dependencies step. 

> 4. Should I add an option to use a socket cluster on Mac/Linux for users who don’t want to fork from within Rstudio?

I guess that makes sense if forking is indeed dangerous in RStudio. I've been using `parallel::mclapply` on Mac for years in RStudio without incident and we've never had anyone ask for that option in our other packages (e.g. loo) that use `parallel::mclapply`, but I'm not opposed to adding it just in case. That said, I don't think we necessarily need to add that before merging this PR. Since we're still in beta we can just wait and see if that extra argument is wanted/needed and then decide before going to CRAN.

jgabry:
> 1. I’ve added documentation of the cores argument in `summarise_draws.R` (see line 21), but I cannot figure out how to get this to show up in the docs at `draws_summary.html`.  Pointers appreciated.

If I do
* `devtools::document()`
* `devtools::load_all()` or reinstall it 
* `?draws_summary`

then I see `cores` documented. Does that work for you?

jsocolar:
> then I see cores documented. Does that work for you?
I see the documentation with `?draws_summary`, but it doesn't show up for me in `docs/reference/draws_summary.html`.  Seems like I'm missing a step to render the html doc.

jgabry:
> I see the documentation with `?draws_summary`, but it doesn't show up for me in `docs/reference/draws_summary.html`. Seems like I'm missing a step to render the html doc.

Oh I see what you mean. That's done using `pkgdown::build_reference()`, although I think we've only been regenerating the website when bumping the version number. That said, we should consider bumping the version number after adding this feature.

R/summarise_draws.R
```diff
@@ -133,47 +189,29 @@ summarise_draws.draws <- function(x, ..., .args = list()) {
...
-  create_summary_list <- function(x, v) {
-    draws <- drop_dims_or_classes(x[, , v], dims = 3, reset_class = FALSE)
-    args <- c(list(draws), .args)
-    v_summary <- named_list(names(funs))
-    for (m in names(funs)) {
-      v_summary[[m]] <- do.call(funs[[m]], args)
+  if (cores == 1) {
+    out <- summarise_draws_helper(x, funs, .args)
+  } else {
+    x <- x[ , , !(dimnames(x)$variable %in% reserved_variables())]
```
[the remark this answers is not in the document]
jsocolar: Not quite:
```
n <- 2
test_array <- array(data = rnorm(1000*nc*n), dim = c(1000,nc,n))
x <- as_draws_array(test_array)
dimnames(x)$variable[2] <- reserved_variables()[1]
length(variables(x))
length(dimnames(x)$variable)
```

jsocolar:
Ok, this is now up-to-date and ready for further review.  Passes R-cmd-check.  All of @jgabry's comments addressed.

I of course have not gone ahead and updated the html documentation; I'll leave that (and decisions about bumping the version) to others.

R/summarise_draws.R
```diff
@@ -60,14 +64,66 @@ summarize_draws <- summarise_draws
...
 #' @export
-summarise_draws.default <- function(x, ...) {
-  x <- as_draws(x)
-  summarise_draws(x, ...)
+summarise_draws.default <- function(.x, ...) {
+  .x <- as_draws(.x)
+  summarise_draws(.x, ...)
+}
+
+# Helper functions
```
jgabry, line 72: Can you move all internal non-exported functions below the exported functions to make it easier to read through the file (someone only needs to look at the internal functions if they really want to understand the guts)

R/summarise_draws.R
```diff
@@ -60,14 +64,66 @@ summarize_draws <- summarise_draws
...
+  for (m in names(funs)) {
+    v_summary[[m]] <- do.call(funs[[m]], args)
+  }
+  v_summary
+}
+
+summarise_draws_helper <- function(.x, funs, .args) {
+  variables_x <- variables(.x)
+  # get length and output names, calculated on the first variable
+  out1 <- create_summary_list(.x, variables_x[1], funs, .args)
```
jgabry: here there's `out1` (no underscore) but just above there's `variables_x` (with underscore). Either convention is fine but I would pick one or the other.

R/summarise_draws.R
```diff
@@ -118,62 +174,59 @@ summarise_draws.draws <- function(x, ..., .args = list()) {
...
-  create_summary_list <- function(x, v) {
-    draws <- drop_dims_or_classes(x[, , v], dims = 3, reset_class = FALSE)
-    args <- c(list(draws), .args)
-    v_summary <- named_list(names(funs))
-    for (m in names(funs)) {
-      v_summary[[m]] <- do.call(funs[[m]], args)
+  if (cores == 1) {
+    out <- summarise_draws_helper(.x, funs, .args)
+  } else {
+    .x <- .x[ , , !(dimnames(.x)$variable %in% reserved_variables())]
```
jgabry: `variables.draws_array()` removes reserved variable names internally. So I think this can be simplified to: 

```r
x <- x[ , , variables(x)]
```

R/summarise_draws.R
```diff
@@ -118,62 +174,59 @@ summarise_draws.draws <- function(x, ..., .args = list()) {
...
-  if (length(variables) > 1L) {
-    for (v_ind in 2:length(variables)) {
-      out_v <- create_summary_list(x, variables[v_ind])
-      out[v_ind, ] <- unlist(out_v)
+    if (checkmate::test_os("windows")) {
+      cl <- parallel::makePSOCKcluster(cores)
+      on.exit(parallel::stopCluster(cl))
+      # Get necessary package functions, including internals, out to the cluster
+      parallel::clusterExport(cl = cl, 
+                              unclass(lsf.str(envir = asNamespace("posterior"), all = T)),
```
jgabry, line 213:
```suggestion
                              unclass(lsf.str(envir = asNamespace("posterior"), all = TRUE)),
```

R/summarise_draws.R
```diff
@@ -118,62 +174,59 @@ summarise_draws.draws <- function(x, ..., .args = list()) {
...
+                              envir = as.environment(asNamespace("posterior"))
+                              )
+      parallel::clusterExport(cl = cl, 
+                              unclass(lsf.str(envir = asNamespace("checkmate"), all = T)),
```
jgabry, line 217:
```suggestion
                              unclass(lsf.str(envir = asNamespace("checkmate"), all = TRUE)),
```

R/summarise_draws.R
```diff
@@ -118,62 +174,59 @@ summarise_draws.draws <- function(x, ..., .args = list()) {
...
+                              envir = as.environment(asNamespace("checkmate"))
+      )
+      parallel::clusterExport(cl = cl, 
+                              unclass(lsf.str(envir = asNamespace("rlang"), all = T)),
```
jgabry, line 221:
```suggestion
                              unclass(lsf.str(envir = asNamespace("rlang"), all = TRUE)),
```

jgabry (commented):
This looks great, thank you @jsocolar. Now that I've looked through it again I have a few more minor style comments.

jgabry:
Aside from the minor style comments, I have a question about the Windows cluster stuff. It looks like the solution ended up being to export _all_ (exported and non-exported) functions from posterior, rlang, and checkmate onto the cluster. That's a clever solution (and I certainly wouldn't have come up with anything better on Windows myself) but there has to be a simpler way, right? @rok-cesnovar or @paul-buerkner (or anyone who knows windows much better than me), is there an alternative to this?

R/summarise_draws.R
jsocolar: So just checking, if true then presumably we can just remove this entire line?

jgabry: If you want to be 100% sure reserved variable names are dropped then I think you still need it: 

```r
x <- example_draws() 
variables(x)[1] <- ".log_weight" # reserved name
print(x, max_variables = 1) # includes variable with reserved name 
print(x[,, variables(x)], max_variables = 1)   # doesn't include variable with reserved name
```

But now I'm wondering why we don't prevent printing reserved variables but want to prevent summarizing them (maybe because we don't usually care about inference on something like `.log_weight`?). @paul-buerkner what do you think?

jsocolar:
@jgabry The alternative is to explicitly list by name all of the functions that we need, including non-exported functions that are called internally to the functions that we're using.  I was worried that doing so would add maintenance burden if anything changes, including if, say, `checkmate` changes the naming or usage of their internal functions (if the exported functions we use use any of their internals).  
That's why I went for this "kitchen sink" approach, which is supposed to replace, e.g. a `library(checkmate)` executed on the cluster, which won't pass checks.

jgabry:
I see what you mean about the cluster stuff. 

> Or maybe there's a better alternative that I don't know about...

If there is I don't know about it either ;) Maybe one of our Windows literate collaborators will enlighten us!

jgabry:
@jsocolar Is it ok if we add you as a Contributor in the DESCRIPTION file? Given the work that went into this and the previous PR speeding up summarise_draws I think we should do that if you'd like to be listed.

jsocolar:
> @jsocolar Is it ok if we add you as a Contributor in the DESCRIPTION file? Given the work that went into this and the previous PR speeding up summarise_draws I think we should do that if you'd like to be listed.

Yeah absolutely.

> I see what you mean about the cluster stuff.

I guess one option would be to reimplement the functions we need from `rlang` and `checkmate` ourselves.  Then we'd just be exporting our own namespace. Haven't checked the licenses for those packages or thought at all about the legal aspect.

jgabry:
Ok great, thanks again for working on this! I just pushed a commit your branch adding you to the DESCRIPTION file. Feel free to edit it if I spelled anything wrong, you want to add a middle initial, or anything else. 

> I guess one option would be to reimplement the functions we need from `rlang` and `checkmate` ourselves. Then we'd just be exporting our own namespace. Haven't checked the licenses for those packages or thought at all about the legal aspect.

Don't worry about it for now. Let's see if anyone has another alternative, but if not I think I'd rather just do what you're already doing now than reimplement all the functions from the other packages.

mjskay:
I believe (if I'm understanding the problem correctly) that the future package uses the [globals](https://cran.r-project.org/web/packages/globals/index.html) package to solve these kinds of problems. Dunno if another dependency is what you wanted as the solution though :).

R/summarise_draws.R
jsocolar: I've implemented @jgabry's simplification here.  I won't resolve this conversation pending comments on @jgabry's comment above.

paul-buerkner: Let's move this discussion to a separate issue.  @jgabry do you want to open it?

paul-buerkner:
I will check the windows stuff now.

paul-buerkner:
I have made some changes to simplify parallelization on Windows using the `globals` package (thanks @mjskay), which is now under suggests. All test pass for me locally on Windows. Let's see what CI checks say. 

I also renamed `cores` to `.cores` to not clash with names of `...` arguments. For the `.x`, I think @jgabry wanted to have a separate discussion, which it fine for me. Once CI checks pass, I am happy with this PR and we can merge on my side. 

@jsocolar Please also check my changes to make sure I didn't change anything to worse.

paul-buerkner:
Ok, so that is interesting. Checks run for me locally on windows but fail on github actions with the same errors @jsocolar got earlier. I will keep investigating.

paul-buerkner:
Now I am also starting my own chain of commits to fix windows issues on github actions :-D 

Let's see how it goes

jsocolar:
I feel your pain @paul-buerkner :(

paul-buerkner:
After some failures, I decided to move back to the approach of @jsocolar. There might be better ways but since I cannot even reproduce the test failure on my own windows machine, I also a little helpless here :-D

Anyway, I approve this PR as is. @jgabry do you want to take another look and merge, if you are fine with it as well?

R/summarise_draws.R
jgabry: Here's the issue: https://github.com/stan-dev/posterior/issues/146

tests/testthat/test-summarise_draws.R
```diff
@@ -68,3 +68,68 @@ test_that("summarise_draws warns if all variable names are reserved", {
   variables(x) <- ".log_weight"
   expect_warning(summarize_draws(x), "no variables with unreserved names")
 })
+
+test_that(paste(
+  "multicore summarise_draws is identical to single-core summarise_draws",
+  "including if some chunks contain no variables"
+), {
+  skip_on_cran()
```
jgabry, line 76: I'm not 100% sure but I think we can run this on CRAN as long as `.cores <= 2`

jsocolar:
Given that the main use case is for really big draws arrays, should we consider warning on 
```
memuse::memuse(object.size(x)) > (memuse::Sys.meminfo()$totalram / (.cores + 1))`
```
or alternatively
```
object.size(x) > (benchmarkme::get_ram() / (.cores + 1))
```
or something similar?  

Dunno if it's worth the extra dependency.  

If we do something like this, what we really should do is set an additional argument that prevents the operation by default unless there's excess RAM.  It could still be overridden by a user who thinks their swap will work quickly enough, but it would prevent a naive user from fitting a huge model that takes forever, then running `summarise_draws` without saving, and then being forced to kill R, potentially losing days of computation in a worst-case scenario (the temp directory is now gone).  So the default could be to not compute, remind the user to save their work, suggest using a lower number of cores, but also tell the user that they can change to `.ram_safety = FALSE` if they want to proceed.

paul-buerkner:
I agree this could be useful, but I think we should merge the current PR without additional features. I am happy to discuss the ideas in a separate issue/PR.

jgabry:
> I agree this could be useful, but I think we should merge the current PR without additional features. I am happy to discuss the ideas in a separate issue/PR.

Was just about to say the same thing. Let's definitely revisit this in a separate discussion.

jgabry:
I made a small change and moved this

```r
if (ndraws(x) == 0L) {
    return(empty_draws_summary())
}
```

up to the top of the function because we don't need to bother with anything else if there are no draws. Otherwise this looks great! If it passes the checks after my latest commit we can go ahead and merge.

jgabry (approved):

jgabry:
Thanks again @jsocolar!

mike-lawrence:
So I failed to check here (or the docs!) before coding a couple different implementations of my own 🤦 But at least I can confirm that @jsocolar's implementation is seemingly the best among the options I tried!

For posterity, here's code to compare serial vs posterior vs multidplyr vs clustermq:
```r
# choose method to test ----

#options:
#	'serial'
#		- regular ol' serial compute
#	'multidplyr'
#		- uses https://github.com/tidyverse/multidplyr
#	'clustermq'
#		- uses https://github.com/mschubert/clustermq
#		- has nice progress bar
#		- supposed to be low-overhead
#		- seemingly crash-prone
#	'posterior'
#		- uses the parallelism implemented directly in posterior

#uncomment desired option below:
method = (
	  'serial'     # times: 109s, 110s
	  # 'multidplyr' # times: #53s, 52s
	  # 'clustermq'  # times: 53s, 55s
	  # 'posterior'    # times: 38s, 38s, 39s *winner*
)

# libs & user defined functions ----
library(tidyverse)
library(magrittr)
system_time = function(x){print(system.time(x))}

# get_par_dxs: function to check parameter diagnostics (rhat, _ess)
get_par_dxs = function(x,nchains){
	x = matrix(x,ncol=nchains)
	tibble::tibble(
		rhat = posterior::rhat(x)
		, ess_bulk = posterior::ess_bulk(x)
		, ess_tail = posterior::ess_tail(x)
	)
}

# importing posterior samples ----
#   n.b. aria::coda() yields a tibble of the same format as the cmdstan csvs
if(!file.exists('post.qs')){
	(
		aria::coda("sampled_2way.qs")
		%>% filter(!warmup)
		%>% select(
			-warmup
			, -ends_with('__')
		)
		%>% select(
			chain
			, iteration
			, starts_with('z_')
			, starts_with(fixed('r.'))
		)
		%>% pivot_longer(
			cols = c(-chain,-iteration)
			, names_to = 'variable'
		)
		%>% group_by(variable)
	) ->
		post
	qs::qsave(post,'post.qs',preset='fast')
}else{
	post = qs::qread('post.qs')
}

# run the chosen method ----
if(method == 'serial'){
	system_time(
		(
			post
			%>% summarise(
				get_par_dxs(
					x = value
					, nchains = length(unique(chain))
				)
			)
		) ->
			par_dxs
	)
}

if(method == 'multidplyr'){
	system_time(cluster <- multidplyr::new_cluster(parallel::detectCores()/2-1))
	system_time(multidplyr::cluster_assign(cluster, get_par_dxs = get_par_dxs))
	system_time(
		post %<>% multidplyr::partition(cluster)
	)
	system_time(
		(
			post
			%>% summarise(
				get_par_dxs(
					x = value
					, nchains = length(unique(chain))
				)
			)
		) ->
			par_dxs
	)
	system_time( par_dxs %<>% collect() )
}

if(method == 'clustermq'){
	options(clustermq.scheduler = "multicore")
	system_time(
		post %<>% group_split()
	)
	system_time(
		(
			clustermq::Q(
				x = post
				, fun = function(x){
					z = get_par_dxs(
						x$value
						,nchains = length(unique(x$chain))
					)
					z$variable = x$variable[1]
					return(z)
				}
				, n_jobs = parallel::detectCores()/2-1
				# , chunk_size = floor(length(post)/(parallel::detectCores()/2-1)/10)
			)
		) ->
			par_dxs
	)
	system_time(
		par_dxs %<>% bind_rows()
	)
}

if(method=='posterior'){
	#posterior wants a wide format
	(
		post
		%>% pivot_wider(names_from=variable)
		%>% rename(
			.chain = chain
			, .iteration = iteration
		)
		%>% posterior::as_draws()
	) -> post
	system_time(
		(
			post
			%>% posterior::summarise_draws(
				posterior::default_convergence_measures()
				, .cores = parallel::detectCores()/2-1
			)
		) -> par_dxs
	)
}

```

Issue #146, opened 2021-05-20
When to avoid reserved variables


jgabry:
Right now we allow printing variables with reserved names
but we drop these variables for methods like `summarise_draws()`. 

```r
x <- example_draws() 
variables(x)[1] <- ".log_weight" # reserved name
print(x, max_variables = 1) # includes variable with reserved name
summarise_draws(x) # doesn't include variable with reserved name
```

Is this the behavior we want?

paul-buerkner:
`print.draws_df` has a `reserved` argument defaulting to FALSE, so I suggest we add this to other print methods as well. The reason, we had this only in `draws_df` was the initially only this format had reserved variables. I will make a PR to fix this issue.

jgabry:
Thanks!

PR #147, opened 2021-05-21
(don't) print reserved variables correctly

paul-buerkner:
This PR should fix issue #146. @jgabry Woud you mind reviewing it?


jgabry (approved):
Looks good! I added some tests for this to `test-print.R`. If it passes the checks after that let's go ahead and merge.

Issue #61, opened 2020-02-18
Coding matrix/vector variables in tidy form


statscats:
When variables are vector/matrix they are stored in a tibble named as a[i,j].

@jgabry  and I were talking about whether there was some utility of creating a convenience function that splits the indices in long form i.e.,

param idx1  idx2
a           i         j

It would be useful for getting things in the right form for Bayesplot.

mjskay:
Unless I'm misunderstanding that is pretty close to what spread_draws / gather_draws are for in tidybayes (amongst other things).

lauken13:
Sorry for the late reply -  I apparently had logged onto a different Github account. 

Yeah that looks perfect! Do you think it's worth wrapping it to jump from the matrix form to this in one step?

mjskay:
Yeah! Long term I'd like to replace `tidybayes::tidy_draws()` with (or have it be a wrapper around) `posterior::as_draws_df()`. The other functions in tidybayes use `tidy_draws()` internally, so once that change happens those other functions should work on any of the formats supported by posterior.

A related question: is there particular functionality in gather/spread_draws that could be abstracted out or which it might be desirable to have exposed in a form that is easier to program against? Currently those functions take in specs for what transformation you want in a non-standard evaluation form, convert them internally to a simpler spec format, then do all the ugly data munging using those. This makes for a nice UI for interactive use but might be a pain to program against. If there's a need/desire for a more programmable variant that shouldn't be a hard change.

The second question is if some of that moves to posterior or not. This might be related to bigger questions of "what to do with tidybayes" --- some time back I mentioned to Jonah that I was contemplating splitting off the geoms from tidybayes into a separate package (maybe "ggdist"). It has evolved from a package just about munging posteriors into a sort of two-headed beast with a bunch of uncertainty vis geoms and stats.

paul-buerkner:
One way of differentiating between tidybayes and posterior functions could be whether data or ploting is involved in addition to posterior draws. So whatever only handles draws could (potentially) go into posterior, but anything including data (such as `add_predicted_draws` etc.) or plotting would be out of scope. Not sure if this is a reasonable threshold though and I don't have a strong opinion here.

lauken13:
Yeah definitely don't want to feature overlap here!

mjskay:
[...]

Issue #43, opened 2019-11-08
summary with one line per _named_ parameter


jgabry:
The discussion in #32 reminded me that one thing @andrewgelman and others have been wanting for a while (and that I think would also be useful as an alternative to the standard summary output) is a summary where each vector/matrix/array parameter only occupies a single line just like scalars. That is, each line corresponds to a named parameter rather than a parameter element. 

(This could be its own thing or just an option to a custom print method for our existing summary objects.)

The quantities to display are debatable (e.g. maybe min of all the individual elements’ ESS, max of the Rhats, etc), so it would be good to get input from Andrew and anyone else interested.

jgabry:
One possibility is something along these lines

```r
variable   min_mean max_mean max_rhat min_ess_bulk min_ess_tail
alpha[1]       4.5     4.5    1.0         881         300
tau[1]         3.8     3.8    1.1         386         311
omega[9]       1.5     7.4    1.0         552         272
Sigma[7,7]    -1.7     5.1    1.0         421         202
```

where the numbers inside brackets indicate the dimensions (we could also omit if a scalar instead of putting `[1]`). Not sure which columns we include, but a concise table like this would be a really nice option for models with many parameters. 

Anyone else have any thoughts on this?

avehtari:
Since brackets are used for indexing, this can cause confusion for R users. Would it be too much to have a column `dim` as the second column?
```
variable dim min_mean max_mean max_rhat min_ess_bulk min_ess_tail
alpha    1       4.5     4.5    1.0         881         300
tau      1       3.8     3.8    1.1         386         311
omega    9       1.5     7.4    1.0         552         272
Sigma    7,7    -1.7     5.1    1.0         421         202
```

paul-buerkner:
dim separately may be a little ugly and not not too helpful for filtering I would say as the structure would vary across variables, that is, sometimes 1 number sometimes 2 or more separated with commas.

jgabry:
What if we just replace the brackets to a different symbol to refer to dimension? There are lots of possibilities. Here are just a few: 

```
variable   min_mean max_mean max_rhat min_ess_bulk min_ess_tail
alpha (1)       4.5     4.5    1.0         881         300
tau   (1)       3.8     3.8    1.1         386         311
omega (9)       1.5     7.4    1.0         552         272
Sigma (7,7)    -1.7     5.1    1.0         421         202
```

```
variable   min_mean max_mean max_rhat min_ess_bulk min_ess_tail
[...]

PR #152, opened 2021-05-21
Roll up summaries of nonscalar variables

jsocolar:
See https://github.com/stan-dev/posterior/issues/43

This is not ready for primetime, but I want the authors to be able to commit here if they want, so here it is.

Still needs:
* Proper error handling
* Tests
* Review of `get_indices.R` and, if it looks good, rewriting `as_draws_rvars.draws_matrix()` to rely on `get_indices` (see also #61)


jsocolar:
Having a think about how to handle `NA` summaries, and I'm not sure what the best approach is.  Some variables will have `NA`s that can be safely ignored.  For example, rhat for Cholesky-factors and the like.  But in other cases, the `NA`s can be crucial diagnostic tools.  For example, if there's a problem in the model definition and a parameter that is supposed to vary is fixed, or if numerical issues cause the sampler to freeze for a single element of a vector.  Since there's no good way to tell whether an `NA` is expected or problematic, we need to decide what to do. Some options:
* Always display `NA` for the maximum or minimum value if any elements are `NA`.  This is the current behavior.
* Provide an argument to ignore `NA`s.
* Ignore `NA`s by default, but if the summary contains any of `ess_bulk`, `ess_tail`, or `rhat`, append an additional column giving the number of elements that are `NA` for any of these quantities.
* Something else...

R/parse_variable_indices.R
```diff
@@ -0,0 +1,148 @@
...
+#'    
+#'    $position  the position of each combination of indices from $indices in the the argument `x`
+#' 
+#' @details
+#' Assumes that variable indexing uses square brackets in the variable names
+#' 
+NULL
+
+parse_variable_indices <- function(x){
+  vars_indices <- strsplit(x, "(\\[|\\])")
```
mjskay, line 35: You use a number of regexes in a row here to parse `x`. Can you combine them into a single regex and use [capture groups](https://stackoverflow.com/a/16563900) to pull out the relevant pieces? That should be faster, and will also be more flexible if at some point in the future we want to provide people with options to parse different formats of indices (like different separators or whatever).

R/parse_variable_indices.R
```diff
@@ -0,0 +1,148 @@
...
+  vars <- sapply(vars_indices, `[[`, 1)
+  var_names <- unique(vars)
+  # Check that no variables contain unpaired or non-terminal square brackets
+  if_indexed <- (sapply(vars_indices, length) > 1)
```
mjskay, line 39: `sapply(x, length)` can be rewritten as `lengths(x)` and it will be faster (`lengths()` is vectorized).

R/parse_variable_indices.R
```diff
@@ -0,0 +1,148 @@
...
+  if_indexed2 <- grepl("\\[.*\\]$", x)
+  extra_bracket <- grepl("\\].", x) | grepl("\\[.*\\[", x)
+  if (any(extra_bracket) | (!identical(if_indexed, if_indexed2))) {
+    stop_no_call(paste("Some variable names contain unpaired square brackets",
+                  "or are multi-indexed."))
+  }
+  missing_index <- grepl("\\[,|,\\]|,,", x) # check for `[,` `,,` or `,]`
```
mjskay, line 46: I would roll this into the regex above instead of making a specific check for it. Given that in most use cases these variable names are auto-generated by model code, I doubt that missing indices will be common enough to warrant time spent on an extra check rather than just having the index fail to parse. (Others might disagree though)

R/parse_variable_indices.R
```diff
@@ -0,0 +1,148 @@
...
+  if (any(missing_index)) {
+    stop_no_call(paste("Some variables contain missing indices. Each comma between square", 
+                       "brackets must be both preceeded and succeeded by an index."))
+  }
+
+  # Get dimensionality. Variables with no brackets are given as dimensionality zero.
+  # Variables with brackets are given dimensionality 1, even if they contain just one element.
+  dimensionality_elementwise <- sapply(vars_indices, function(x){
+    if (length(x) == 2) {
+      out <- length(strsplit(x[2], ",")[[1]]) # number of commas plus one
```
mjskay, line 56: Why not generate the `dimensionality_elementwise` variable later, after you have parsed the variable indices? By generating it here you have to parse the indices twice; generating it later you could parse them just once.

R/parse_variable_indices.R
```diff
@@ -0,0 +1,148 @@
...
+    out
+  })
+  dimensionality <- sapply(var_names, function(x){
+    out <- unique(dimensionality_elementwise[vars == x])
+    if (length(out) != 1) {
+      stop_no_call(paste0("Inconsistent indexing found for variable ", x, " ."))
+    }
+    out
+  })
+  variable_indices_info <- lapply(var_names, function (x) {
```
mjskay, line 69: You have three instances of *apply over the same structure (`var_name`) in a row here. Is each element of the intermediate variables (`dimensionality[[i]]`, `dimensionality_elementwise[[i]]`) only used in corresponding element `i` of this last apply? If so I would skip creating these intermediate variables and move the relevant code into the final apply. This would also resolve my previous comment.

R/parse_variable_indices.R
```diff
@@ -0,0 +1,148 @@
...
+    # position gives the position in the original vector of variables (passed to `parse_variable_indices`)
+    # where a given combination of indices is found (NA if the comination does not exist)
+    indices_info <- named_list(c("dimensionality", "dimensions", "implied_dimensions", "index_names", "indices", "position"))
+    indices_info$dimensionality <- dimensionality[[x]]
+    var_i <- vars == x
+    var_length <- sum(var_i)
+    if (dimensionality[x] == 0) {
+      # single variable, no indices
+      indices_info$dimensions <- indices_info$implied_dimensions <- NA
+      indices_info$indices <- as.integer(1)
```
mjskay, line 82: `as.integer(1)` can be written as an integer literal: `1L`
```suggestion
      indices_info$indices <- 1L
```

R/parse_variable_indices.R
```diff
@@ -0,0 +1,148 @@
...
+      dimnames(indices_info$indices) <- NULL
```
mjskay, line 83: dimnames of a newly-created integer should already be NULL, I think you can omit this line.
```suggestion
```

R/parse_variable_indices.R
```diff
@@ -0,0 +1,148 @@
...
+      indices_info$dimensions <- unname(apply(indices, 2, function(x){length(unique(x))}))
+      unique_indices <- vector("list", length(indices))
+      .dimnames <- vector("list", length(indices))
+      names(unique_indices) <- names(indices)
+      for (i in seq_along(indices)) {
+        numeric_index <- suppressWarnings(as.numeric(indices[[i]]))
+        if (!anyNA(numeric_index) && rlang::is_integerish(numeric_index)) {
+          # for integer indices, we need to convert them to integers
+          # so that we can sort them in numerical order (not string order)
+          if (min(numeric_index) >= 1) {
```
mjskay, line 99: I would factor out the common code from inside both branch of this if statement and put it before it (in both branches the result should be converted to an integer):

```r
indices[[i]] <- as.integer(numeric_index)
```

R/parse_variable_indices.R
```diff
@@ -0,0 +1,148 @@
...
+          factor_levels <- unique(indices[[i]])
+          indices[[i]] <- factor(indices[[i]], levels = factor_levels)
+          # these aren't sorted so they appear in original order
+          unique_indices[[i]] <- factor(factor_levels, levels = factor_levels)
+          .dimnames[[i]] <- unique_indices[[i]]
+        }
+      }
+      
+      indices_info$index_names <- .dimnames
+      indices_info$implied_dimensions <- unname(sapply(unique_indices, length))
```
mjskay, line 120: As above, I would use `lengths()` here
```suggestion
      indices_info$implied_dimensions <- unname(lengths(unique_indices))
```

R/parse_variable_indices.R
```diff
@@ -0,0 +1,148 @@
...
+      # here because it helps us do the sort after the merge, where
+      # we need to sort in reverse order of the indices (because
+      # the value of the last index should move slowest)
+      all_indices <- expand.grid(rev(unique_indices))
+      # merge with all.x = TRUE (left join) to fill in missing cells with NA
+      indices <- merge(all_indices, cbind(indices, index = seq_len(nrow(indices))),
+                       all.x = TRUE, sort = FALSE)
+      # need to do the sort manually after merge because when sort = TRUE, merge
+      # sorts factors as if they were strings, and we need factors to be sorted as factors
+      indices <- indices[do.call(order, as.list(indices[, -ncol(indices), drop = FALSE])),]
```
mjskay, line 139: Even though I wrote this sort originally I'm not actually sure this sort is needed, because I think the `all_indices` grid might already be sorted in the correct order? worth checking on this, because if so dropping this sort will be a big time saver on variables with lots of indices.

R/parse_variable_indices.R
```diff
@@ -0,0 +1,148 @@
...
+#' @name variable_indices
+#' @param x a character vector of variables
+#'
+#' @return
+#' A list with index information for each unique variable name V in `x`. Top-level list names are 
+#' the variable names. Each element contains: 
+#'    $dimensionality  the number of dimensions of V. Returns 0 for scalars with no brackets
+#'    but 1 for `y[1]` even if `y` has no other entries in `x`.
+#'    
+#'    $dimensions  a vector of the actual dimensions of V, as determined by the number of unique
```
mjskay, line 13: Is `dimensions` currently used for anything?

Or put another way: is there a situation in which it is clear that someone would ever want both `dimensions` and `implied_dimensions` at the same time? I asked because I think it could simplify things to just calculate and return the implied dimensions as the canonical "dimensions". Possibly with an option to only calculate the non-implied dimensions if those are desired (in which case the function can skip all the expensive calculations involved in figuring out the implied dimensions).

mjskay:
> Review of parse_variable_indices.R and, if it looks good, rewriting as_draws_rvars.draws_matrix() to rely on parse_variable_indices

Thanks @jsocolar! I added (hopefully not too many :) ) comments with some suggestions for simplification. It's very exciting to see this getting done, it's gonna be a very useful function (I am probably going to rewrite some tidybayes stuff on top of it :)).

A minor point of discussion about the format: currently you have `dimensionality` and `dimensions` elements. Normally I'm all for fully spelling things out, but in this case I might suggest naming these `ndim` and `dim` respectively. That would make people's code that uses this function slightly more self-documenting, as those names are used frequently throughout R code to correspond to the two concepts these variables represent.

R/parse_variable_indices.R
jsocolar: I think the main use case for both would be if somebody wants to check whether they match.  Not sure that's a "real" use case though.  

In `rollup_summary` I thought it might be more useful to report the `dimensions` rather than the `implied_dimensions` since the former give a better (albeit imperfect) sense of big the rolled-up variable actually is.

TBH, for my own purposes I don't expect `dimensions` and `implied_dimensions` to ever differ, so I'll definitely follow your lead on which quantity is more useful.

mjskay: > I think the main use case for both would be if somebody wants to check whether they match. Not sure that's a "real" use case though.

Yeah I'd vote for waiting to see if this ever comes up instead of designing for it in advance.

> In rollup_summary I thought it might be more useful to report the dimensions rather than the implied_dimensions since the former give a better (albeit imperfect) sense of big the rolled-up variable actually is.

Good point, I agree. In that case, that function doesn't need any of the stuff that comes with implied_dimensions right? Then I would propose adding a flag to parse_variable_indices that either fills in the implied dimensions or not (with default `TRUE`). I'm not sure what to call it -- `fill_implied_dimensions` is too long. Maybe just `fill` or `fill_dim` or `fill_indices` or `fill_missing`? Then if `fill_missing` is `TRUE`, `dimensions` would be reported with missing indices filled in and otherwise not.

R/parse_variable_indices.R
jsocolar: I could well be missing a clever trick here, but I think this and the previous comment are not quite straightforward.

Of these three apply statements, the first applies a function over `vars_indices` and the second and third apply over `var_names`.  It would be easy to bring the second inside the third.  However, the second apply statement is very cheap and includes a check for inconsistent dimensionality.  I thought it might be better to error quickly on inconsistent dimensionality rather than start the relatively expensive variable parsing.

R/parse_variable_indices.R
jsocolar: done

R/parse_variable_indices.R
mjskay: Ah I think I misunderstood what this was doing. 

I think this still could be simplified so that the parsing of variable indices only has to be done once (which should be faster and also have lower maintenance cost, since we wouldn't have to modify that parsing in two places for future features). If you move [this line](https://github.com/stan-dev/posterior/pull/152/files#diff-27e20aebb5dcbf1421dc3aaa68d69b3ebc556f813dadadff19d5814951f70cd6R81):

```r
      indices <- sapply(vars_indices[var_i], `[[`, 2)
```

Above its containing if/else block, then I think you could also factor out the vectorized strsplit from the line after the above one to get a list of indices, upon which you could then use lengths / etc to get elementwise ndims and check for consistency. Then index parsing only happens once / in one place.

R/parse_variable_indices.R
jsocolar: done

R/parse_variable_indices.R
mjskay: To expand on this: here's a regular expression that matches well-formed indexed variable names. Starting/ending the regex with "^" and "$" guarantees it must match the whole name, so failures imply an ill-formed name:

```r
x <- c("a[1]", "b[", "c[1,2]", "d[]", "e]", "f")
m <- regexec("^([^][]+)(\\[([^][]+)\\])?$", x)
vars_matches <- regmatches(x, m)
vars_matches
```
```
[[1]]
[1] "a[1]" "a"    "[1]"  "1"   

[[2]]
character(0)

[[3]]
[1] "c[1,2]" "c"      "[1,2]"  "1,2"   

[[4]]
character(0)

[[5]]
character(0)

[[6]]
[1] "f" "f" ""  "" 
```

Then you can check for ill-formed names with something like this:

```r
if (any(lengths(vars_matches) == 0)) {
   stop_no_call("insert informative error")
}
```

Variables without indices will have the fourth element of their corresponding entry in `vars_matches` be empty, and variables with indices will have it be the indices which can then be split.

This regex is fairly strict, and I'm not sure how strict we want to be --- e.g., do we want names like `"x]"` to fail, or do we want them just to be parsed as non-indexed variables? I can see an argument for a looser version of this parsing that rarely (or maybe never) fails but instead lets weird-looking names go by and treats them as non-indexed. E.g. something like this:

```r
x <- c("a[1]", "b[", "c[1,2]", "d[]", "e]", "f")
m <- regexec("^(.+?)(\\[(.+)\\])?$", x)
vars_matches <- regmatches(x, m)
vars_matches
```
```
[[1]]
[1] "a[1]" "a"    "[1]"  "1"   

[[2]]
[1] "b[" "b[" ""   ""  

[[3]]
[1] "c[1,2]" "c"      "[1,2]"  "1,2"   

[[4]]
[1] "d[]" "d[]" ""    ""   

[[5]]
[1] "e]" "e]" ""   ""  

[[6]]
[1] "f" "f" ""  "" 
```

i.e. in this example `"b["`, `"d[]"`, and `"e]"` all get treated as weird-looking variable names.

This is also a bit more flexible in that it would be pretty easy to allow people to pass in different choices for the opening/closing bracket and update the regex accordingly.

jsocolar:
@mjskay just mentioning that I have some time to jump back on this now.  Once I remember where we're at, I'll keep it moving along.

mjskay:
Wonderful to hear! I think that the variable index parsing aspects of this PR may also end up being related to some of the conversation in #187 about exposing internal functions for flattening / unflattening arrays. I think that work can be done after this PR though, since it's already doing a couple of different things at once.

PR #153, opened 2021-05-22
Convergence functions for rvars

mjskay:
This PR will address one the last remaining big pieces of #117, which is allowing convergence functions to be applied directly to rvars. I have drafted this just with support for `ess_bulk()` at the moment so I can get feedback on the approach before applying it to the other convergence functions. @paul-buerkner and @jgabry both of your feedback would be appreciated on this. **If you think the approach I have taken with `ess_bulk()` makes sense**, I will go ahead and apply it to the other functions.

The basic idea is that convergence functions can be applied directly to rvars and they will behave like other summary functions applied to rvars do. For example:

```r
d = as_draws_rvars(example_draws("multi_normal"))
summarise_draws(d)
## # A tibble: 12 x 10
##    variable      mean  median    sd   mad     q5     q95  rhat ess_bulk ess_tail
##    <chr>        <dbl>   <dbl> <dbl> <dbl>  <dbl>   <dbl> <dbl>    <dbl>    <dbl>
##  1 mu[1]       0.0514  0.0575 0.112 0.131 -0.130  0.225  1.01      677.     356.
##  2 mu[2]       0.111   0.104  0.199 0.198 -0.208  0.449  1.00      566.     426.
##  3 mu[3]       0.186   0.184  0.314 0.315 -0.322  0.715  1.02      650.     334.
##  4 Sigma[1,1]  1.28    1.26   0.165 0.173  1.03   1.56   1.00      742.     369.
##  5 Sigma[2,1]  0.525   0.502  0.200 0.173  0.227  0.874  1.01      454.     239.
##  6 Sigma[3,1] -0.403  -0.393  0.282 0.267 -0.874  0.0432 1.01      468.     308.
##  7 Sigma[1,2]  0.525   0.502  0.200 0.173  0.227  0.874  1.01      454.     239.
##  8 Sigma[2,2]  3.67    3.62   0.447 0.433  3.02   4.40   1.01      529.     363.
##  9 Sigma[3,2] -2.10   -2.11   0.480 0.469 -2.87  -1.39   1.02      434.     357.
## 10 Sigma[1,3] -0.403  -0.393  0.282 0.267 -0.874  0.0432 1.01      468.     308.
## 11 Sigma[2,3] -2.10   -2.11   0.480 0.469 -2.87  -1.39   1.02      434.     357.
## 12 Sigma[3,3]  8.12    8.02   0.946 0.941  6.71   9.91   0.997     729.     324.
ess_bulk(d$Sigma)
##          [,1]     [,2]     [,3]
## [1,] 742.2907 454.0657 468.3890
## [2,] 454.0657 528.7972 434.1141
## [3,] 468.3890 434.1141 728.9440
```

Under the hood this would involve making each convergence diagnostic function generic and providing an implementation for rvars, as I have done for ess_bulk here: https://github.com/stan-dev/posterior/compare/rvar-conv?expand=1#diff-2cb156b102d99b6f7f88a258bf8225d63d8310a5681181ae93956a6a71cb94b2R96-R106

I think/hope that is a relatively lightweight change to make to each diagnostic function. I also updated the documentation templates for convergence diagnostics, which it would be helpful to get feedback on.

The other changes in this PR are:
- The addition of a `with_chains` argument to `rvar()` and `draws_of()` that allows people to optionally get/set the draws of an rvar using an array of shape iterations x chains x other dims instead of just draws x other dims. I realized something like this was needed (at least on the retrieval side) for doing the convergence functions and figured it would be easy enough to expose to users in case it was useful for them. I still think the draws x other dims format is the right default format for input/output/storage/etc, but this change makes it easier for people to choose in cases where it would help to have chains in the array itself.
- An unrelated minor change I was working on for #117 is some bugfixes for as.data.frame.rvar and as_tibble.rvar


paul-buerkner:
The design and example implementation look good to me!

jgabry:
Yeah I like this approach too. 

> I think/hope that is a relatively lightweight change to make to each diagnostic function. 

The only complication I can think (not so much because it's difficult but because it requires a decision) is how to handle diagnostic functions that return multiple values. For example, what would   `mcse_quantile(d$Sigma, probs = c(0.25, 0.75))` return? A list of two 3x3 matrices, one for each element of `probs`? Something else? I'm not sure what makes the most sense. 

> * The addition of a `with_chains`

I also think this makes sense. As part of this PR can you add examples to the `rvar` and `draws_of` doc demonstrating using the new `with_chains` argument?

mjskay:
> The only complication I can think of (not so much because it's difficult but because it requires a decision) is how to handle diagnostic functions that return multiple values. For example, what would   `mcse_quantile(d$Sigma, probs = c(0.25, 0.75))` return? A list of two 3x3 matrices, one for each element of `probs`? Something else? I'm not sure what makes the most sense. 

Yeah me neither. Currently my plan was to follow the pattern I used for `range.rvar()`, which uses the first dimension of the returned array. E.g.:

```r
d = as_draws_rvars(example_draws("multi_normal"))
d$mu
## rvar<100,4>[3] mean ± sd:
## [1] 0.051 ± 0.11  0.111 ± 0.20  0.186 ± 0.31 
range(d$mu)
##            [,1]       [,2]       [,3]
## [1,] -0.3292973 -0.5178862 -0.6442688
## [2,]  0.3643947  0.6350847  0.9797682
```

So mcse_quantile would do this:

```r
mcse_quantile(d$mu)
##                 [,1]       [,2]       [,3]
## mcse_q5  0.012675014 0.01279547 0.02331627
## mcse_q95 0.008302994 0.02609606 0.05555146
```

But this format is basically what I chose because I didn't have strong opinions, and it is what `apply()` happened to give me. I don't have a strong use case in mind for this so I don't have much else to go on, but if others have a good reason to change it we should. The other format I considered was putting the extra dimension at the end of the array. I suppose putting it at the beginning is at least consistent with where we put the draws dimension.

If we decide to change this I will also update range.rvar() and such to match.

> I also think this makes sense. As part of this PR can you add examples to the rvar and draws_of doc demonstrating using the new with_chains argument?

Oh yeah, will do.

jgabry:
> Currently my plan was to follow the pattern I used for `range.rvar()`, which uses the first dimension of the returned array

That seems reasonable. I don't have a better idea at the moment.

paul-buerkner:
I agree.

mjskay:
Alright, I think this is ready for review now. 

(N.B. the Github CI failure on macOS devel appears due to failures in the setup of R-devel not related to this PR)

R/convergence.R
```diff
@@ -43,14 +43,25 @@ NULL
 #' mu <- extract_variable_matrix(example_draws(), "mu")
 #' rhat_basic(mu)
 #'
+#' d <- as_draws_rvars(example_draws("multi_normal"))
+#' rhat_basic(d$Sigma)
+#'
+#' @export
+rhat_basic <- function(x, split = TRUE) UseMethod("rhat_basic")
```
paul-buerkner, line 50: Since we are now making a lot of these functions generic, I think we should talk about the argument structure of the generic. Most notably:

- Should the generics have a `...` argument to be more extensible by other packages?
- Should additional arguments such as `split` already be part of the generic, or should we rather go with the most generel `x, ...` signature?

paul-buerkner (commented):
Looks good to me. I have only one general discussion question about generic argument signatures we should decided.

R/convergence.R
mjskay: Good point. I think adding `...` makes sense for sure. I'm not sure which other arguments are necessary in the base generics, it seems like this might vary depending on the function?

jgabry: > Good point. I think adding ... makes sense for sure. I'm not sure which other arguments are necessary in the base generics, it seems like this might vary depending on the function?

I agree that adding `...` makes sense and that for other arguments it would depend on the function. For example, for `ess_quantile` 

```r
ess_quantile(x, probs = c(0.05, 0.95), names = TRUE)
```

I think it could make sense for `probs` to be in the generic because it's an essential argument but the `names` argument probably doesn't need to be part of the generic. That said, I'd also be fine with just going with the more general `x,...` signature and leaving all the other arguments for the methods.

paul-buerkner:
Adding `probs` to the quantile generrics sounds good! Any other suggestions for arguments in the generics? Right now, I couldn't think of one at least.

jgabry:
> Adding `probs` to the quantile generrics sounds good! Any other suggestions for arguments in the generics? Right now, I couldn't think of one at least.

Yeah adding probs to the generics for `ess_quantile` and `mcse_quantile` is the only thing I can think of.

jgabry:
> > Adding `probs` to the quantile generrics sounds good! Any other suggestions for arguments in the generics? Right now, I couldn't think of one at least.
> 
> Yeah adding probs to the generics for `ess_quantile` and `mcse_quantile` is the only thing I can think of.

Aside from this I think this PR is ready to go!

mjskay:
great! I'll update the generics per this conversation (probably won't get to this until later next week)

mjskay:
Okay, I made the changes based on the above conversation:

- I added `...` to all convergence generics
- I kept `probs` in convergence generics with a `probs` argument
- I also kept `na.rm` since that seemed sensible to keep (but I can remove if folks disagree).
- I removed `split` and `names` from several generics as neither seemed to be mentioned by folks as desirable in the generic (do we still agree with this? I could see an argument for leaving `split` and `names` in their generics). This required changing the argument order for quantile2.default from `(x, probs, names, na.rm, ...)` to `(x, probs, na.rm, names, ...)` since the first three args are in the generic but `names` is not.

Thoughts?

paul-buerkner:
Thank you! Looks good to me. I am indifferent with respect to `na.rm`. Both is fine for me. If @jgabry approves as well, I think we can merge it.

mjskay:
A quick thing here that came up in #166 is that we don't have an example of `with_chains` in the rvar vignette. Since it's in this PR I will quickly add it to the vignette here before we merge.

mjskay:
Okay, minor change to this PR: I added examples of `with_chains = TRUE` and `draws_rvars()` (to contrast it with `as_draws_rvars()`) to the rvar vignette. @paul-buerkner let me know if those look good.

paul-buerkner:
Looks good to me!

paul-buerkner:
Actually, I think we can merge right now. @mjskay any additional changes you want to make?

mjskay:
I'm good, merge away!

PR #154, opened 2021-05-22
Types in parentheses in doc

### LINKED ISSUE #144: summarise_draws: can't use x as function name
I haven't had a chance to look into this yet, but an error is thrown when attempting to use `x` as the name in `summarise_draws()`. Every other name I've tried works: 

```r
draws <- example_draws()
summarise_draws(draws, y = mean) # fine
summarise_draws(draws, x = mean) # error
Error: Don't know how to transform an object of class 'function' to any supported draws format.
```

### LINKED ISSUE #150: Adding type information more explicitly to the doc
In the CmdStanR documentation @rok-cesnovar and I introduced the convention of clearly indicating in parentheses the object type that each argument is expecting (see examples below). This is very atypical for R packages but I think greatly improves the documentation and I've gotten really positive feedback on it from users. Ultimately I'd like to convert the other stan-dev R packages to this style if there's support for it. 

Would you all be open to using this style for posterior? If not no worries, but if so then I'm happy to be the one to implement it. 

---

Here are examples of what I'm referring to from CmdStanR:

If there's just one option we just put it in parentheses: 

```r
#' @param chains (positive integer) The number of Markov chains to run. 
```

If there are multiple options we put "multiple options" in parentheses and then use markdown syntax to create an itemized list of the options: 

```r
#' @param data (multiple options) The data to use for the variables specified in
#'   the data block of the Stan program. One of the following:
#'  * A named list of \R objects (like for RStan). Internally this list is then
#'  written to JSON for CmdStan using [write_stan_json()].
#'  * A path to a data file compatible with CmdStan (JSON or \R dump). See the
#'  appendices in the CmdStan manual for details on using these formats.
#'  * `NULL` or an empty list if the Stan program has no data block.
```

jgabry:
Closes #150 by adding the expected type in parentheses when documenting function arguments. 

A few notes: 

* For now I only added the types in parentheses in the doc for exported functions not internal functions 
* The only arguments that don't have the type in parentheses are the `...` arguments. If you see any other arguments without it that’s an oversight on my part.
* I went ahead and changed the `x` to `.x` in `summarise_draws()`, which also closes #144


paul-buerkner:
Looks good to me. Merging now.

Issue #155, opened 2021-05-23
Should we rename `draws_rvars()` as `draws_rvar()`?


Ozan147:
All sub-classes of `draws` has the `draws_{parent}` singular naming structure except for `draws_rvars`. Is this intentional to stress that `draws_rvar` can hold multiple `rvar` objects? The singular version still seems more consistent to me.

> array -> draws_array
> df -> draws_df
> list -> draws_list
> matrix -> draws_matrix
> rvar -> draws_rvar

paul-buerkner:
I would tend to disagree on this. The reason we call it `draws_matrix`, `draws_array` etc. is that those are actually valid objects of parent class `matrix`, `array`, etc. However, `draws_rvar` does not have `rvar` as parent class but is actually a **list** of `rvar`s. Hence the name `draws_rvars`. What do others, in particular @mjskay, think?

mjskay:
Yes, I agree with @paul-buerkner's logic, which reflects our original discussion of this issue where we considered both names. I would add that I read "draws_x" as "adjective_noun" which makes sense for "draws_rvars" but not for "draws_rvar" because the noun in this case is not one rvar but multiple rvars.

Ozan147:
Sounds good, it makes sense. Sorry for bringing up already discussed/resolved questions 😅

paul-buerkner:
No worries, we very much appreciate your work and thoughts you put into posterior!

Issue #115, opened 2021-03-24
as_draws() does not work correctly on mcmc.list objects even though it is supported by more specific as_draws_...() functions


mjskay:
It appears that `as_draws()` in selecting an appropriate target draws format "XXX" for an object bypasses S3 dispatch on the chosen `as_draws_XXX()` function, which has the slightly strange (from a user perspective) result that `as_draws(x)` may not be the same as `as_draws_XXX(x)` even if "XXX" is the format chosen by `as_draws()`. 

For example, here is the (incorrect) translation of an `"mcmc.list"` object into a draws object produced by `as_draws()`:

```r
data(line, package = "coda")
as_draws(line)
```
```
# A draws_list: 600 iterations, 1 chains, and 2 variables

[chain = 1]
$line1
 [1] 7.2 3.0 3.7 3.3 3.7 3.6 2.7 3.0 3.5 2.1

$line2
 [1] 2.1 2.6 3.7 2.9 3.1 3.0 2.4 3.3 2.7 3.0

# ... with 590 more iterations
```

And the correct output given by `as_draws_list()`:

```r
as_draws_list(line)
```
```
# A draws_list: 200 iterations, 2 chains, and 3 variables

[chain = 1]
$alpha
 [1] 7.2 3.0 3.7 3.3 3.7 3.6 2.7 3.0 3.5 2.1

$beta
 [1] -1.57  1.50  0.63  1.18  0.49  0.21  0.88  1.09  1.07  1.48

$sigma
 [1] 11.23  4.89  1.40  0.66  1.36  1.04  1.29  0.46  0.63  0.91


[chain = 2]
$alpha
 [1] 2.1 2.6 3.7 2.9 3.1 3.0 2.4 3.3 2.7 3.0

$beta
 [1] 0.95 0.53 0.52 0.98 0.94 0.66 0.52 0.94 1.06 0.83

$sigma
 [1] 2.85 1.71 0.79 0.95 0.90 0.35 1.09 0.97 0.69 0.64

# ... with 190 more iterations
```

The problem appears to stem from this location:

https://github.com/stan-dev/posterior/blob/d0845b3f606a09c51da73e03a6a3ebe2b3c1389d/R/as_draws.R#L47-L52

[...]

PR #157, opened 2021-05-23
Restructured class transformation functionality - `as_draws_*()`

Ozan147:
Addresses #115 ~~and #155.~~

This suggestion restructures the `as_draws_*()` methods and functions in order to:

- resolve existing transformation issues mentioned in #115 
- condense and simplify transformation methods as much as possible
- avoid deep nested lists and loops, and vectorize the process as much as possible

~~Additionally, it introduces the singularization of `rvar` related names by converting `draws_rvars` to `draw_rvar` in accordance with #155.~~

This PR is completely experimental. We can merge only some parts or none of it. I did not intend to make it this big but all these functions were closely connected and one thing led to another. Let me know what you think so that I can iterate.


R/as_draws_df.R
```diff
@@ -90,47 +59,7 @@ as_draws_df.draws_array <- function(x, ...) {
...
-  as_draws_df(as_draws_array(x), ...)
-}
-
-#' Convert any \R object into a `draws_df` object
-#' @param x An \R object.
-#' @param .iteration optional name of the column containing iteration indices
-#' @param .chain optional name of the column containing chain indices
-#' @noRd
-.as_draws_df <- function(x, .iteration = NULL, .chain = NULL) {
+as_draws_df.data.frame <- function(x, .iteration = NULL, .chain = NULL, ...) {
```
Ozan147, line 62: Just a head-up; I realized I didn't pay much attention to ellipsis (...) in this PR. Depending on your answer to #158, I'll go through the code and make sure they are placed when necessary.

mjskay:
Thanks for this!

With respect to performance-oriented changes, have you benchmarked or profiled the new implementations? I find using something like [profvis](https://rstudio.github.io/profvis/) can be very helpful in identifying where to focus performance-enhancing efforts. When I benchmark your new implementation it seems to be as slow as (or slower than) the existing implementation on most things I try.

E.g. consider this object with a variable with a bunch of character indices (the use of character indices here doesn't seem to affect speed really, I was using mostly to test edge cases):

```r
iterations = 1000
chains = 4
I = 50
J = 50
x = as_draws_array(array(1:(iterations*chains*I*J), dim = c(iterations, chains, I*J), dimnames = list(
  NULL, NULL, paste0("x[", rep(as.roman(1:I), each = J), ",", rep(as.roman(1:J), I), "]")
)))
x
```
```
# A draws_array: 1000 iterations, 4 chains, and 2500 variables
, , variable = x[I,I]

         chain
iteration 1    2    3    4
        1 1 1001 2001 3001
        2 2 1002 2002 3002
        3 3 1003 2003 3003
        4 4 1004 2004 3004
        5 5 1005 2005 3005

, , variable = x[I,II]

         chain
iteration    1    2    3    4
        1 4001 5001 6001 7001
        2 4002 5002 6002 7002
        3 4003 5003 6003 7003
        4 4004 5004 6004 7004
        5 4005 5005 6005 7005

, , variable = x[I,III]

         chain
iteration    1    2     3     4
        1 8001 9001 10001 11001
        2 8002 9002 10002 11002
        3 8003 9003 10003 11003
        4 8004 9004 10004 11004
        5 8005 9005 10005 11005

, , variable = x[I,IV]

         chain
iteration     1     2     3     4
        1 12001 13001 14001 15001
        2 12002 13002 14002 15002
        3 12003 13003 14003 15003
        4 12004 13004 14004 15004
        5 12005 13005 14005 15005

# ... with 995 more iterations, and 2496 more variables
```

Benchmarking conversions from this to other formats on master, I get this:

```r
bench::mark(check = FALSE, min_iterations = 20, filter_gc = FALSE,
  as_draws_df(x),
  as_draws_rvars(x),
  as_draws_matrix(x)
)
```
```
# A tibble: 3 x 13
  expression              min  median `itr/sec` mem_alloc `gc/sec` n_itr  n_gc total_time result memory   
  <bch:expr>         <bch:tm> <bch:t>     <dbl> <bch:byt>    <dbl> <int> <dbl>   <bch:tm> <list> <list>   
1 as_draws_df(x)      202.1ms 230.7ms      3.42     307MB     8.72    20    51      5.85s <NULL> <Rprofme~
2 as_draws_rvars(x)    93.1ms 106.6ms      6.59     231MB    10.5     20    32      3.04s <NULL> <Rprofme~
3 as_draws_matrix(x)   19.7us  21.3us  43750.          0B     4.38 10000     1   228.57ms <NULL> <Rprofme~
# ... with 2 more variables: time <list>, gc <list>
```

And on your implementation I get:

```
# A tibble: 3 x 13
  expression            min median `itr/sec` mem_alloc `gc/sec` n_itr  n_gc total_time result memory time 
  <bch:expr>         <bch:> <bch:>     <dbl> <bch:byt>    <dbl> <int> <dbl>   <bch:tm> <list> <list> <lis>
1 as_draws_df(x)      197ms  296ms      3.36     345MB     9.25    20    55      5.95s <NULL> <Rpro~ <ben~
2 as_draws_rvars(x)   226ms  305ms      2.99     348MB     8.37    20    56       6.7s <NULL> <Rpro~ <ben~
3 as_draws_matrix(x)  225ms  241ms      3.59     345MB     7.71    20    43      5.58s <NULL> <Rpro~ <ben~
# ... with 1 more variable: gc <list>
```

I think maybe it would help to have a set of benchmarks we can evaluate against. I'll open an issue for this.

Ozan147:
Thanks for the feedback! Yep, this implementation seems to be slower than the current version on most transformations. This is mostly because I forced all conversions to go through `data.frame`s, which is costly and not ideal (e.g. `draws_array` <-> `draws_matrix` conversion is straightforward, no need to stop by `draws_df`). I just wanted to have an experimental PR up to get some 👀  on it, and will be iterating to find better transformations in each case.

mjskay:
> Thanks for the feedback! Yep, this implementation seems to be slower than the current version on most transformations. This is mostly because I forced all conversions to go through data.frames, which is costly and not ideal (e.g. draws_array <-> draws_matrix conversion is straightforward, no need to stop by draws_df). I just wanted to have an experimental PR up to get some 👀 on it, and will be iterating to find better transformations in each case.

Cool! Can you say a bit more about the high-level strategy for the restructuring? I think it would help in discussing what direction to go.

For example, my understanding of the current structure (@paul-buerkner can correct me if I'm wrong) is to initially route calls through `as_draws()` so it can try to find the "closest" format of a given type. This (in theory) makes it possible for any type to be translated to any other, hopefully efficiently (given smart choices for "closest", in many cases this should be efficient). The wrinkle came in where `as_draws()` could not take advantage of S3 dispatch on finding the method to do the conversion since that might cause infinite recursion (e.g. if `as_draws_list.default()` called `as_draws()` and it called back to `as_draws_list.default()` we would have a problem). This (I believe) is why `as_draws()` calls to an internal `.as_draws_XXX()` instead.

My suggestion in #115 was to use reflection to solve this problem. Around line 50 here:

https://github.com/stan-dev/posterior/blob/d0845b3f606a09c51da73e03a6a3ebe2b3c1389d/R/as_draws.R#L47-L52

You could insert a check to see if a non-default method is defined for the class in question, and if so dispatch to that instead of the `.as_draws_XXX()` method (so can see a function for that kind of check here: https://github.com/mjskay/tidybayes/compare/posterior#diff-b531cc0ee8d06f169b52f7fd1c86fd6b5cc6abc960b6a0121ff21af71ac32b88R53-R62). This should avoid infinite recursion while still making it easy to provide specialized, efficient implementations of conversions between types.

I'm asking because I'm trying to evaluate pros/cons of code maintenance given the different potential implementation strategies.

Ozan147:
Declining this PR as it's not mature enough and out-of-date.

PR #160, opened 2021-05-24
Continuous benchmarking

mjskay:
This is an attempt at using the development version of bench to do continuous benchmarking. I have no idea if we want to merge something like this in at this time, but it seems to more-or-less work so I thought I'd see what folks think. I'm very open to suggestions on how to make this process work better.

I added a README about in `bench/README.Rmd`. Basic idea is that a Github action will run benchmarks and store them in the [git notes](https://git-scm.com/docs/git-notes) attached to the commit. You can then use `bench::cb_fetch()` to grab benchmarks and plot them. You can also run them yourself with `bench::cb_run()` locally. 

Benchmarks are stored in the `bench/` folder. I added one example benchmarking the performance of converting draws objects between the various formats (this benchmark can probably be improved). 

There isn't any automated reporting at the moment, but I figured that for now folks can just add relevant plots to the README and run it manually when doing benchmarking-related work. Currently it just outputs this one plot:

image

which doesn't show anything meaningful at the moment since those two benchmarks were run on different machines.

You do have to install the github version of bench to run the `cb_XXX()` functions. I did not make this a requirement on the package via `Remotes:`, because CRAN does not allow that field; instead the github action manually grabs the github version of bench.

### DIFF
--- a/.Rbuildignore
+++ b/.Rbuildignore
@@ -10,3 +10,4 @@
 ^docs$
 ^pkgdown$
 ^\.github$
+^bench$
--- a/.github/workflows/continuous-benchmarks.yaml
+++ b/.github/workflows/continuous-benchmarks.yaml
@@ -1,0 +1,31 @@
+on: push
+
+name: Continuous Benchmarks
+
+jobs:
+  build:
+    runs-on: macOS-latest
+    steps:
+      - name: Checkout repo
+        uses: actions/checkout@master
+
+      - name: Setup R
+        uses: r-lib/actions/setup-r@master
+
+      - name: Install dependencies
+        run: |
+          Rscript -e "install.packages(c('remotes'))" -e "remotes::install_deps(dependencies = TRUE)"
+          Rscript -e "remotes::install_github("r-lib/bench")"
+          R CMD INSTALL .
+
+      - name: Fetch existing benchmarks
+        run: Rscript -e 'bench::cb_fetch()'
+
+      - name: Run benchmarks
+        run: Rscript -e 'bench::cb_run()'
+
+      - name: Show benchmarks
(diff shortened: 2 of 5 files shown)


jgabry:
> I have no idea if we want to merge something like this in at this time, but it seems to more-or-less work so I thought I'd see what folks think. I'm very open to suggestions on how to make this process work better.

This is great. I haven't yet had a chance to think much about this particular process, but I'm definitely in favor of adding this sort of benchmarking functionality. 

> You do have to install the github version of bench to run the `cb_XXX()` functions. I did not make this a requirement on the package via `Remotes:`, because CRAN does not allow that field; instead the github action manually grabs the github version of bench.

I think that's fine.

mjskay:
Now that I've played with this more, I'm not convinced the approach bench is using will be directly usable. It records benchmark results attached to a git ref but without recording the branch that was at that ref at the time, which makes comparisons a bit annoying. Plus it doesn't auto-generate PR messages, which I think would be useful.

I came across another experimental package that does CI benchmarking differently: when a PR is made, it runs benchmarks on both the PR branch and the target in the same CI run and reports relative changes as a message on the PR. That seems a better "out-of-the-box" fit. See here: https://lorenzwalthert.github.io/touchstone/index.html

It's a relatively new and experimental package, but it's written by and being used by the maintainer of styler in that package, so it's maybe not a big risk to try it.

jgabry:
Yeah you’re right, that package does seem like a better fit. It also sounds like it’s likely to be developed and maintained well and even if it isn’t it wont break anything in posterior from a user perspective.

mjskay:
Closing this for now until I (or someone else) has time to play with touchstone.

Issue #156, opened 2021-05-23
Chain and iteration info for `draws_matrix`


Ozan147:
All `draw_*` objects include the chain and iteration details except for `draws_matrix`. Would it make sense to find a way to augment `draws_matrix` with this info, perhaps by making it more similar to `draws_df`? I personally like that the remaining classes can be transform to each other seamlessly without losing any info. However `draws_matrix` collapses all draws into a single chain, making it impossible to revert or recover other classes with multiple chains.
```r
x_array <- example_draws()
# A draws_array: 100 iterations, 4 chains, and 10 variables

x_df <- as_draws_df(x_array)
# A draws_df: 100 iterations, 4 chains, and 10 variables

x_matrix <- as_draws_matrix(x_df)
# A draws_matrix: 400 draws, and 10 variables

x_df <- as_draws_df(x_matrix)
# A draws_df: 400 iterations, 1 chains, and 10 variables

x_array <- as_draws_array(x_matrix)
# A draws_array: 400 iterations, 1 chains, and 10 variables
```

paul-buerkner:
We had thought about this when starting to develop posterior and decided to explictly let `draws_matrix` be a format without chain information, thus primarily to be used once you have already convinced yourself that your model has converged (or you have 1 chain anyway, or even just MC draws) so you could use matrices for increased convenience compared to 3D arrays. The hazzle it would need to somehow stored the chain information there (e.g., as an attribute) and then make sure this information is carried around without R dropping it is not worth it in my opinion. 

@mjskay went to some lengths to get chains infromation into the `rvar` format, but this is such a custom empty-list-with-attributes format (even though nobody sees that from the outside) that we have way more control to drop/keep stuff in certain operations anyway. However, even there we had been debating on just leaving it out. 

Accordingly, in my opinion, we should leave it as is, but I am happy to hear other opinions and be convinced otherwise.

mjskay:
It's an interesting question. I would suspect that in most cases if you want to use an array-like format but keep chain information, it probably makes the most sense to just use draws_array instead of draws_matrix. I think the big value of draws_matrix is either (like @paul-buerkner says) when you don't care about chains anymore or when you are trying to get output that's already in that format into a draws format.

jgabry:
[...]

PR #170, opened 2021-06-26
support chains in `draws_matrix' format

paul-buerkner:
This PR adds chain supprt to the `draws_matrix` format as discussed in #156.

@mjskay as you already have experience with chain support via attributes, would you mind reviewing this PR?


R/as_draws_matrix.R
```diff
@@ -148,6 +163,9 @@ is_draws_matrix_like <- function(x) {
   out <- NextMethod("[", drop = drop)
   if (length(dim(out)) == length(dim(x))) {
     class(out) <- class(x)
+    if (missing(i)) {
```
mjskay, line 166: Do we want a warning message here in the case where `i` is not missing and `nchains(x) > 1`? Currently chains are silently dropped:

```r
> xm = as_draws_matrix(example_draws())
> xm
# A draws_matrix: 100 iterations, 4 chains, and 10 variables
    variable
draw   mu tau theta[1] theta[2] theta[3] theta[4] theta[5] theta[6]
  1  2.01 2.8     3.96    0.271    -0.74      2.1    0.923      1.7
  2  1.46 7.0     0.12   -0.069     0.95      7.3   -0.062     11.3
  3  5.81 9.7    21.25   14.931     1.83      1.4    0.531      7.2
  4  6.85 4.8    14.70    8.586     2.67      4.4    4.758      8.1
  5  1.81 2.8     5.96    1.156     3.11      2.0    0.769      4.7
  6  3.84 4.1     5.76    9.909    -1.00      5.3    5.889     -1.7
  7  5.47 4.0     4.03    4.151    10.15      6.6    3.741     -2.2
  8  1.20 1.5    -0.28    1.846     0.47      4.3    1.467      3.3
  9  0.15 3.9     1.81    0.661     0.86      4.5   -1.025      1.1
  10 7.17 1.8     6.08    8.102     7.68      5.6    7.106      8.5
# ... with 390 more draws, and 2 more variables
> xm[c(1,101,201,301), ]
# A draws_matrix: 4 iterations, 1 chains, and 10 variables
     variable
draw   mu tau theta[1] theta[2] theta[3] theta[4] theta[5] theta[6]
  1   2.0 2.8      4.0     0.27    -0.74     2.10     0.92     1.65
  101 3.0 2.8      6.3     1.03     0.22    -0.73     4.06     3.34
  201 1.8 8.7     13.3     2.11     1.38     7.82    -1.50    12.74
  301 6.5 3.8      5.8     4.96     5.69     4.84     2.74     0.61
# ... with 2 more variables
```

E.g. rvar currently tries to do something like this when it needs to merge chains:

```r
> x1 = rvar(rnorm(100))
> x2 = rvar(rnorm(100), nchains = 4)
> x1 + x2
rvar<100>[1] mean ± sd:
[1] 0.019 ± 1.6 
Warning message:
Random variables do not have the same number of chains (1 and 4),
so chains were dropped.
Use merge_chains() to collapse chains before combining rvars with
a different number of chains to avoid this warning. 
```

Something similar might work in this case. The only caveat might be if it is annoying to people who expect matrices not to yell at them. On the other hand, people could always just do `merge_chains()` on the matrix first to avoid the error.

mjskay:
This looks great! I had only the one minor question about whether or not we need a warning when slicing causes chains to be merged. I don't have a strong opinion on which way to go with that.

R/as_draws_matrix.R
paul-buerkner: Thank you for reviewing this PR! Personally, I would prefer to not have a warning in this case as it requires users to use `merge_chains` every time they use the `draws_matrix` as a matrix. I can imagine this warning to be too annoying overall, although of course I understand its uses. Would you be ok with not including the warning for now? We could still add it later on if we think its inclusion is required. 

Alternatively, we can throw the warning only if a certain option is set to true, kind of a safe mode that yells at users whenever the kill chain information with their transformations? If you think that could be sensible, how shall we call such an option?

mjskay: Yeah, to be honest I have found the chain-merging warning annoying in rvar as well :). 

I would be in support of an option to turn on or off both warnings (maybe "posterior.warn_on_merge_chains"?)

paul-buerkner: Sounds good. And we would default it to `FALSE` then?

mjskay: Sure, fine with me.

paul-buerkner: Good. I will add this to the PR and also edit the `rvar` operations to use this option.

mjskay: Thanks!

paul-buerkner: Done. Would you mind taking another look?

mjskay: Looks great! The only last thing I would say is we should probably add a description of the option under "Package options" in posterior-package.R.

jgabry:
Thanks for adding this!

Issue #159, opened 2021-05-23
Performance benchmarks


mjskay:
Pinging off of https://github.com/stan-dev/posterior/pull/157#issuecomment-846611937, would it be useful to have some standard benchmarks we use for evaluating performance?

Does anyone have experience setting something like this up? I noticed that the [bench](https://github.com/r-lib/bench) package looks to be developing infrastructure for doing this with continuous integration using scripts with calls to `bench::mark()` in a `bench/` subdirectory. Might be worth playing with.

jgabry:
Good idea. I haven't used bench but at first glance I agree it seems worth playing with. Could be a good solution.

mjskay:
Noting here so we don't forget: from the conversation over at https://github.com/stan-dev/posterior/pull/160#issuecomment-847435468 I think that [touchstone](https://lorenzwalthert.github.io/touchstone/index.html), not bench, is the more promising direction for this.

assignUser:
Hey @mjskay! As discussed {touchstone} is now in a stable state, would love your feedback!

PR #171, opened 2021-06-27
"magic dispatch" for as_draws.default()

mjskay:
This PR implements the "magic dispatch" fix for #115 by allowing `as_draws.default(x)` to dispatch to `as_draws_XXX(x)` instead of `.as_draws_XXX(x)` when we know there is an implementation of `as_draws_XXX` for the class of `x`.

As compared to @Ozan147's approach in #157 it is a more minimal approach to solving the problem for the purposes of a CRAN release. That said, I think both approaches should ultimately be compatible with each other: particularly, I think there may be transformations between specific draws classes that have faster implementations in #157. It's possible we could combine these approaches by adding those specific faster transformations incrementally without the need for larger restructuring. This would also be aided by having performance benchmarks (#159).

I would be curious what folks think, particularly @Ozan147 since you've thought about this problem.


paul-buerkner:
The PR looks good from my side! Thanks for handling this! Anyone still wanting to check it out? I will leave it open for a day or two and merge it then, if there are no objections.

paul-buerkner:
I will merge this now.

Issue #165, opened 2021-06-23
CRAN release


jgabry:
I think posterior is getting close to the point where we should consider releasing the first CRAN version. I'm opening this issue so we have a place to discuss the things we need/want to get done before then.

mjskay:
I am 100% on board with this, as I already have feature branches of ggdist and tidybayes that I can't submit to cran because they depend on posterior :).

Is it worth asking: is there anything missing from the [CRAN milestone](https://github.com/stan-dev/posterior/milestone/3)? Or conversely, anything on there that isn't necessary for the first CRAN release? e.g. nice-to-haves that wouldn't change the API in a backwards-incompatible way.

paul-buerkner:
I agree. I think one minor thing would be to decide what to do with the `iter` argument in `subset_draws.draws_matrix` if we want to add chain information to the format later on. My guess would be that it's best right now to disallow this argument for now and require users to always subset by `draw` which will remain valid even after adding chain information. 

With regard to the CRAN milestone, I think issues #42, #77, and #105 could be dropped from the milestone and also handled later. The other issues we could try to fix more or less right away when somebody has time.

jgabry:
> Is it worth asking: is there anything missing from the [CRAN milestone](https://github.com/stan-dev/posterior/milestone/3)?

I think we should decide whether to include chain info in `draws_matrix`. But I'm a bit biased because my motivation for that is the CmdStanR issue related to losing chain info (stan-dev/cmdstanr#523). I'm more eager than I used to be to decide one way or another on this issue so that we know if we need to implement a somewhat hacky solution in CmdStanR or not. Are you all ok with deciding on this before the CRAN release? 

> With regard to the CRAN milestone, I think issues #42, #77, and #105 could be dropped from the milestone and also handled later.

Agreed. I just went ahead removed them from the milestone (but if anyone disagrees we can add them back).

paul-buerkner:
I agree, we should probably add chain info to the `draws_matrix` format. There are apparently enough use cases for this, to make it worthwhile implementing and maintaining.

paul-buerkner:
With all issues of the CRAN milestone now closed, I think we can actually move towards the submission.

@jgabry @mjskay @avehtari since you are authors of this package, I would like to get your explicit approval for the CRAN submission.

Asking anybody reading this, is there anything that you think needs to be added to the first CRAN release? Specfically things that are much harder to change once things are on CRAN?

jgabry:
[...]

PR #175, opened 2021-07-05
add @return for remaining doc pages except reserved-variables

mjskay:
This PR adds `@return`s for functions docs mentioned in #165, with the exception of `reserved-variables`.


paul-buerkner:
Thanks! Looks good to me.

PR #176, opened 2021-07-05
make rstar example "donttest" instead of "dontrun"

mjskay:
This changes the `rstar()` example from "dontrun" to "donttest", and also wraps it conditionally to run only if {caret} is installed (as {caret} is only in "Suggests"). For #165.


paul-buerkner:
Thanks! Looks good to me.

PR #177, opened 2021-07-05
export reserved_variables and document it

mjskay:
As one possible solution for the problem of `reserved-variables.Rd` not having a documented return value in #165, this PR exports `reserved_variables()`, gives it a `@return` and `@examples` section, and moves the documentation that was in `reserved-variables` to the Details section of `reserved_variables`.

I'm not sure if this is the right solution, but I figured I'd propose it and see what y'all think. I don't know if there was a reason not to export `reserved_variables`, for example.


paul-buerkner:
I think there previously was no need to export `reserved_variables` but I don't see a problem with exporting it either. Thanks for working on this. Merging now.

Issue #164, opened 2021-06-04
rstar() requires {caret} to be loaded manually by the user


mjskay:
I noticed while working on #153 that `rstar()` requires the user to load {caret} manually:

https://github.com/stan-dev/posterior/blob/6cf8d24e8dd97640ec976d0cba56b4fcf69b2d0b/R/rstar.R#L96-L99

Is there a reason why this cannot be done with `requireNamespace()` as is typical (and as recommended in the Writing R Extensions docs)? Making the user load {caret} also loads {lattice} and {ggplot2} into their namespace. 

I briefly tested changing this back to `requireNamespace("caret", quietly = TRUE)` and it seemed to work, but I wasn't sure if I was missing something. @ben18785 it looks like you made this change, is there something that precipitated it?

ben18785:
Thanks @mjskay -- good point. What you propose makes sense I think. So would we want to do something like the below then?

`
if (!requireNamespace("pkg", quietly = TRUE)) { 
    stop_no_call("Package 'caret' is required for 'rstar' to work. Please install this package.")  }
`

mjskay:
Yeah, I think that looks right.

jgabry:
I tried switching to `requireNamespace` but that results in an error when using certain values for the `method` argument: 

```r
# using branch  require-namespace-for-caret 

x <- example_draws("eight_schools")
rstar(x, method = "rf") # works
rstar(x, method = "gbm") # works
rstar(x, method = "knn") # errors unless I do library(caret) first
```

Any idea why some methods don't require `library(caret)` but others do?

ben18785:
This is an idiosyncrasy of the Caret package as I remember. I think the rf
and gbm methods behind the scenes call other packages whereas knn is
natively Caret. For some very odd reason, that's why the latter needs Caret
to be loaded whereas the others don't.

mjskay:
Hm, maybe we should file a bug report on caret?

ben18785:
I just tried:

caret::train(x, y, method="knn")
caret::train(x, y, method="gbm")
caret::train(x, y, method="rf")

and these all worked. So not sure what's going on actually? :)

mjskay:
Huh. Looks like it's something to do with `trainControl()`.

Assuming caret is not loaded:

```r
print(.packages())
```
```
[1] "stats"     "graphics"  "grDevices" "utils"     "datasets"  "methods"   "base"  
```

`method = "rf"` works with `trControl = trainControl(method = "none")`:

[...]

PR #181, opened 2021-07-14
Manually attach/detach caret as needed for rstar

mjskay:
I was thinking about the whole caret namespace thing (#164) and realized there's another way to solve it: to attach and detach caret just for the duration of the `rstar()` function if the user does not have it attached already. This PR implements that solution, which makes us no longer dependent on caret changing how it expects users to use it.

#### Copyright and Licensing

### DIFF
--- a/R/rstar.R
+++ b/R/rstar.R
@@ -95,9 +95,20 @@ rstar <- function(x, split = TRUE,
                   nsimulations = 1000,
                   ...) {
 
-  if (!"caret" %in% .packages()) {
-    stop_no_call("Package 'caret' is required for 'rstar' to work. ",
-          "Please use library(caret) to load the package.")
+  # caret requires itself to be attached to the search list (not just loaded).
+  # To avoid polluting the user's namespace we manually attach caret and its
+  # two hard dependencies (ggplot2 and lattice) if they aren't already attached,
+  # and unload any of them we had to attach when this function exits.
+  loaded_packages <- character()
+  on.exit(for (package in loaded_packages) {
+    detach(paste0("package:", package), character.only = TRUE)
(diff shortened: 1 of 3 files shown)


jgabry:
I like this solution! 

I wonder what CRAN would think about attaching packages listed in Suggests (or not even listed at all like ggplot2), but maybe they're ok with it and/or maybe they won't notice it.

mjskay:
Oh yeah, good point. I took a look at `withr::with_package()`, which does something similar. They call `get("library")(...)` instead of `library(...)`, this could be a way around CRAN noticing undesirable (to them) `library()` calls. We could similarly use `get("require")(...)` here.

I'm also not sure if we should then include `ggplot2` and `lattice` in our `Suggests`? They would be installed by installing caret anyway...

paul-buerkner:
yeah `get("something")` is a way to work around automatic checks of this sort :-D

I would prefer not to have ggplot2 and lattice in suggests if we don't have to, but I am also ok with adding them there if required.

In general I like your idea, and think we can just go ahead and try it out if it passes CRAN checks the next time we release.

From my side, we can merge as is. @jgabry what do you think?

jgabry:
> From my side, we can merge as is. @jgabry what do you think?

Yeah I'm fine with trying this and seeing what CRAN says, or if you want to switch to using get() that's fine too.

paul-buerkner:
@mjskay which of the two versions would you prefer? I don't mind either way.

mjskay:
I'll add the `get()` just to be extra safe, but leave out ggplot and lattice from suggests :).

paul-buerkner:
Great. Merging now.

PR #190, opened 2021-08-12
Use matrixStats to optimize convergence metrics

alyst:
#### Summary

The PR uses `matrixStats::colVars` and base matrix statistics functions (`rowMeans`, `colMeans`) to vectorize the calculation of statistics of matrix rows and column.
That should speed-up a bit and reduce the resource footprint of rhat and ess calculation.

#### Copyright and Licensing


paul-buerkner:
Thank you for this PR. I am a bit overwhelmed at the moment so not sure when I can take a look. Anybody who wants to step in an check this PR?

mjskay:
I don't have much time atm either unfortunately :/. Quick thoughts:

- This does add one dependency (though happily matrixStats has no other dependencies itself). So dependency-wise it seems fine as long as the speedup is worth it.
- Given that, it might help to have some before/after benchmarks to know the speedup.
- Looking at the [list of functions it provides](https://cran.r-project.org/web/packages/matrixStats/vignettes/matrixStats-methods.html), I could see several of these functions also being used to speedup calculations elsewhere, e.g. many of the summary functions for rvars. So if it seems like the speedup is good we could apply this in many places.

alyst:
The [performance benchmarks for *matrixStats* package](https://github.com/HenrikBengtsson/matrixStats/wiki/Benchmark-reports) indicate that using *matrixStats* functions or even the built-in ones (e.g. *colMeans()*) is much faster than *apply() + mean()*.

paul-buerkner:
Thank you for doing additional testing. I did remove `test_draws.rds` in the end in order to keep the package size as small as possible but I  do confirm equivalence on this and some other examples. Merging as soon as checks pass.

paul-buerkner:
@mjskay with this PR merged you can now go ahead and add `matrixStats` functionality to `rvars`.

mjskay:
Great will do! I've already started a branch and I'll try to turn that around soon.

PR #193, opened 2021-08-21
allow as_draws_rvars to accept lists of lists, closes #192

### LINKED ISSUE #192: `as_draws_rvars` on a list of data frames
I am trying to fix issue https://github.com/stan-dev/cmdstanr/issues/536 which is essentially supporting `draws_rvars` directly in cmdstanr and have bumped into a problem for which I have a solution but I dont really like it and am wondering if anything can be done about this in the posterior package.

This is a script that resembles what cmdstanr does internally, read the CSV files to a list of data frames. This is the fastest and most RAM friendly:

```r
library(cmdstanr)
library(posterior)

fit <- cmdstanr_example(example = "logistic")

draws <- list()
i <- 1
for(out_file in fit$output_files()) {
  fread_cmd <- paste0("grep -v '^#' '", out_file, "'")
  draws[[i]] <- data.table::fread(cmd = fread_cmd, data.table = FALSE)
  i <- i + 1
}
```

After this we run the `as_draws_*` on that list (depending on the selected format or as_draws_array by default) and this is currently supported for all formats except for rvars. The problem is that running `as_draws_rvars` directly on this list errors (second line below), while if I first pass it through any of the others, it works (the last line).

```r
> a <- as_draws_list(draws)
> b <- as_draws_rvars(draws) 
Error: Can't convert <data.frame> to <rvar<1>>.
Run `rlang::last_error()` to see where the error occurred.

> c <- as_draws_rvars(as_draws_list(draws))
```

Is there anything that can be done to support this use case directly. Its not a huge deal if we need to pass it through another format, but would be great if we can avoid it. Thanks!

mjskay:
#### Summary

This fixes #192 by falling back to the default list conversion behavior in as_draws_rvars.list() unless the input is a list of rvars.

#### Copyright and Licensing

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,10 +1,10 @@
 Package: posterior
 Title: Tools for Working with Posterior Distributions
-Version: 1.0.1
-Date: 2021-07-14
+Version: 1.0.1.9000
+Date: 2021-08-22
 Authors@R: c(person("Paul-Christian", "Bürkner", email = "paul.buerkner@gmail.com", role = c("aut", "cre")),
              person("Jonah", "Gabry", email = "jsg2201@columbia.edu", role = c("aut")),
-             person("Matthew", "Kay", email = "mjskay@umich.edu", role = c("aut")),
+             person("Matthew", "Kay", email = "mjskay@northwestern.edu", role = c("aut")),
              person("Aki", "Vehtari", email = "Aki.Vehtari@aalto.fi", role = c("aut")),
              person("Måns", "Magnusson", role = c("ctb")),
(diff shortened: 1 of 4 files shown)


paul-buerkner:
Thank you! I think, we need to start documenting our fixes in NEWS and bump the version number so that cmdstanr can target the right version (or later) and be sure the fix is in place. Could you make these changes to the PR?

mjskay:
Sure, done! Dunno if you have a preferred convention for dev versions; I use `x.x.x.9000` as I've seen other packages do that (it generates a NOTE on check which is nice so you don't forget to bump it).

paul-buerkner:
Perfect, thank you!

Issue #197, opened 2021-09-03
New CRAN release


paul-buerkner:
As discussed with @avehtari, the constant-per-chain checks for convergence diagnostics is overly conservative especially for `ess_tail` and creates a lot of confusion. I will make the corresponding change asap and the aim for a CRAN resubmission soon. 

Which other issues do we need to handle before this resubmission?

jgabry:
I looked through the open issues and I don't think there's anything we _must_ have in order for you to do a release (i.e., no major bugs as far as I can tell).

paul-buerkner:
Thank you! Anything you think we should be adding still even if not strictly necessary?

mjskay:
I think #200 is ready and #201 should be shortly as well. I don't see anything else.

paul-buerkner:
With the two PRs merged, I think posterior is now ready for a new CRAN release. I will handle this tomorrow

jgabry:
Thanks Paul!

mjskay:
Yeah, thanks Paul!

PR #200, opened 2021-09-04
Rvar matrix stats

mjskay:
#### Summary

Dunno if you want this before the next CRAN submission, but given discussion in #197 I figured I would submit it now.
This PR follows #190 to use functions from matrixStats to speed up rvar summary functions (in some cases, substantially so). It also adds a NEWS entry for this change and for #190, since it did not have one.

There are also some minor changes to base cases for some of the rvar summary functions that I think make them more consistent. Previously, for some summary functions there was a distinction between `rvar_XXX()` and `rvar_XXX(NULL)`; e.g. `rvar_mean()` would raise an error and `rvar_mean(NULL)` would return `NA`. This was to be consistent with *some* base functions that do this (e.g. `mean()` is an error and `mean(numeric())` is `NA`). Yet some do not return an error (e.g. `min()` returns `Inf` with a warning, as does `min(numeric())`). But the semantics of the rvar summary functions is that they take any number of rvars and collapse them into a single rvar vector before computing the summary, so an empty list and a NULL value should be equivalent; thus, I think the semantics of `min()` are better to mimic. Thus the following changes

- `rvar_mean()`, `rvar_median()`, `rvar_sd()`, `rvar_mad()`, and `rvar_var()` now all return `rvar(NA)` when given either an empty argument list or a `NULL` argument, consistent with (e.g.) `mean(numeric())`, `sd(numeric())`, etc. Also consistent with `sd(numeric())` and others, these functions no longer raise warnings in these cases.
- Following `matrixStats::rowMins()`, `::rowMaxs()`, etc., `rvar_min()`, `rvar_max()`, and `rvar_range()` return `rvar(Inf)`, `rvar(-Inf)`, and `as_rvar(c(Inf, -Inf))` *without* a warning. Previously these returned warnings, but I think these base case values are semantically consistent so don't need warnings per se; this is also the approach taken by the matrixStats functions.

I think these changes make the interface to the summary functions more internally consistent.

#### Copyright and Licensing


tests/testthat/test-summarise_draws.R
```diff
@@ -82,12 +82,12 @@ test_that(paste(
   x <- as_draws_array(test_array)
   sum_x <- summarise_draws(x)
   parsum_x <- summarise_draws(x, .cores = cores)
-  expect_identical(sum_x, parsum_x)
```
mjskay, line 85: I had to change this test and the following one to expect_equal as I was getting "equal but not identical" errors for them on my machine. Not sure why, but I assume this is related to changes to matrixStats impacting parallel execution in some way? In any case I don't think we need to worry if the parallel version produces identical results as long as it is equal down to the default low tolerance?

mjskay:
Also, the codecov decreases here are because: (1) I have left in the `Summary.rvar()` generic in the very unlikely case something calls it in the future (all functions I am aware of that could call it now have faster specific implementations using matrixStats, so there aren't any existing functions that call it) and (2) I have left in `summarise_rvar_within_draws()` even though it is no longer used in favor of `summarise_rvar_within_draws_via_matrix()` as it may be useful in the future for implementing summary functions that do not have matrixStats equivalents.

tests/testthat/test-summarise_draws.R
paul-buerkner: I agree. `expect_identical` is likely too strict in many cases.

mjskay: Great!

paul-buerkner:
Thank you, the PR looks good to me. Merging this now.

PR #201, opened 2021-09-05
Keep variable order in subset_draws()

### LINKED ISSUE #188: subset_draws doesn't respect variable order with arrays
Thanks for this package making Stan easier to use!

This issue is related to #132, where @jgabry noted issues when the requested variables were `c("theta[2]", "theta[1]")`. I'm guessing this issue with arrays was also introduced in #130. I noticed it upgrading from posterior 0.1.2 to 1.0.1. 
The issue in #132 remains fixed.

```r
# before (and expected):
posterior:::check_existing_variables(c("theta", "mu"), posterior::example_draws())
#> [1] "theta[1]" "theta[2]" "theta[3]" "theta[4]" "theta[5]" "theta[6]" "theta[7]"
#> [8] "theta[8]" "mu"

# now:
posterior:::check_existing_variables(c("theta", "mu"), posterior::example_draws())
#> [1] "mu"       "theta[1]" "theta[2]" "theta[3]" "theta[4]" "theta[5]" "theta[6]"
#> [8] "theta[7]" "theta[8]"
```

cc @Ozan147

karldw:
#### Summary

This PR tries to fix #188. The basic approach I took here is to `grep` through the variable names. This is slow, and increases non-linearly in the number of variables, so probably isn't what you actually want to include. Hopefully the test cases are useful.

If you see a way to make this significantly faster, let me know!

I skipped the performance test for now. It takes 11s to run with 10,000 variables (the test requires 300,000 variables in less than a second).

#### Copyright and Licensing

### DIFF
--- a/R/draws-index.R
+++ b/R/draws-index.R
@@ -465,36 +465,29 @@ check_existing_variables <- function(variables, x, regex = FALSE,
   } else if (!scalar_only) {
     # need to find variables that are matched by either a scalar or vector
     # variable in x and what the matching variable is, while keeping original
-    # order of `variables` => we'll do two left joins to fill in scalar and
-    # vector variable names, which will also find missing variables (as NAs)
+    # order of input `variables`
 
-    # because the mapping of scalar variables to variables(x) is one-to-one, can
-    # do the first left join using match instead of merge, which is faster
-    scalar_variables <- all_variables[match(variables, all_variables)]
-
-    # need to use merge for the second join as vector variables may match
-    # multiple variables in `x`
-    all_variables_base <- gsub("\\[.*\\]$", "", all_variables, perl = TRUE)
-    variables_df <- merge(
-      # left join seems to rearrange input order, so need to keep it
-      # and the order of variables in x so we can restore them later
-      data.frame(
-        variable = variables, input_order = seq_along(variables), scalar = scalar_variables,
-        stringsAsFactors = FALSE
-      ),
-      data.frame(
-        variable = all_variables_base, vector = all_variables, variable_order = seq_along(all_variables),
-        stringsAsFactors = FALSE
-      ),
-      by = "variable",
-      sort = FALSE,
-      all.x = TRUE
-    )
-
-    vector_or_scalar <- with(variables_df, ifelse(is.na(vector), scalar, vector))
-    missing_variables <- variables_df$variable[is.na(vector_or_scalar)]
-    variables <- vector_or_scalar[order(variables_df$input_order, variables_df$variable_order)]
-    variables <- variables[!is.na(variables)]
+    # find scalar variables (1-to-1 match between all_variables and variables)
+    scalar_input_ixs <- match(all_variables, variables)
+    # find vector variable matches (match all_variables with the indexing stripped)
+    all_variables_base <- all_variables
+    # exclude already matched scalar variables
+    all_variables_base[!is.na(scalar_input_ixs)] <- NA_character_
+    all_variables_base <- gsub("\\[.*\\]$", "", all_variables_base, perl = TRUE)
+    vector_input_ixs <- match(all_variables_base, variables)
+    # compose the vector of indices of matched input variables
+    input_ixs <- c(scalar_input_ixs[!is.na(scalar_input_ixs)],
+                   vector_input_ixs[!is.na(vector_input_ixs)])
+    # compose the vector of indices of matched all_variables
+    all_var_ixs <- seq_along(all_variables)
+    all_var_matched_ixs <- c(all_var_ixs[!is.na(scalar_input_ixs)],
+                             all_var_ixs[!is.na(vector_input_ixs)])
+    # select missed input variables
+    missing_vars_mask <- rep_len(TRUE, length(variables))
+    missing_vars_mask[input_ixs] <- FALSE
+    missing_variables <- variables[missing_vars_mask]
+    # select matched all_variables maintaining the input variables order
+    variables <- all_variables[all_var_matched_ixs[order(input_ixs, all_var_matched_ixs)]]
   } else {
     missing_variables <- setdiff(variables, all_variables)
   }
--- a/R/draws-index.R
+++ b/R/draws-index.R
@@ -474,7 +474,7 @@ check_existing_variables <- function(variables, x, regex = FALSE,
 
     # need to use merge for the second join as vector variables may match
     # multiple variables in `x`
-    all_variables_base <- gsub("\\[.*\\]$", "", all_variables)
+    all_variables_base <- gsub("\\[.*\\]$", "", all_variables, perl = TRUE)
     variables_df <- merge(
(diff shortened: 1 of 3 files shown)


mjskay:
This is really helpful, thanks! 

Your tests inspired me to make a branch off your PR ([joins-for-188](https://github.com/stan-dev/posterior/compare/joins-for-188)) and take a stab at a faster implementation (see [here](https://github.com/stan-dev/posterior/commit/e359674f4a35ea800428c72b695526594c2d0492)). I realized that the operation in question is basically two left joins: once for input variables onto scalar variable names and then again onto vector variable names. That allows us to identify missing variables and recover the input order in the same set of operations. 

I'm not sure if there's a faster way, but it runs in half a second for 300,000 variables on my machine. That's slower than the implementation on master, which runs in about 0.2 seconds, but of course does not preserve input order. If we're happy with this solution we might therefore need to up the threshold on the performance test slightly.

@karldw since I branched off your branch, maybe the easiest way would be if you merge the joins-for-188 branch into your branch before we accept this PR?

karldw:
Go for it! Before this gets merged, we should remove the `skip("Performance is not tolerable")`

mjskay:
Good point, I removed that skip and converted the PR from draft mode.

R/draws-index.R
```diff
@@ -461,14 +463,38 @@ check_existing_variables <- function(variables, x, regex = FALSE,
...
-    variables <- unique(
-      c(variables[!(variables %in% missing_candidates)],
-        all_variables[all_variables_base %in% vector_variables])
+    variables_df <- merge(
+      # left join seems to rearrange input order, so need to keep it
+      # and the order of variables in x so we can restore them later
+      data.frame(
+        variable = variables, input_order = seq_along(variables), scalar = scalar_variables,
+        stringsAsFactors = FALSE
+      ),
```
alyst, line 484:
```suggestion
      tibble(
        variable = variables, input_order = seq_along(variables), scalar = scalar_variables
      ),
```

Since *tibble* is imported, it's possible to use it for more reliable default behaviour.

paul-buerkner: Does using a tibble affect performance?

mjskay: I just did some profiling and benchmarking and saw no difference in performance for tibble versus data.frame here (I have seen other cases where tibble slowed stuff down, but it seems not here). I have no strong preference in tibble versus data.frame here so it's up to you.

I did notice a slight improvement using perl regexes in the gsub call there, so I changed that to `perl = TRUE`.

alyst: I guess `tibble()` construction is a bit slower because of quosures processing etc, but here's it's just 3 columns, so the impact should be minor, unless this code is used in some heavy loops.

paul-buerkner:
let's stick to data.frame then.

Alexey Stukalov ***@***.***> schrieb am Di., 7. Sept. 2021,
10:24:

R/draws-index.R
alyst: Out of curiosity I've made a merge-free version :)
I'll publish the branch in ~1h. In theory, it should be faster than the current one.

paul-buerkner:
nice! would really like to see the speed differences! thank you!

Alexey Stukalov ***@***.***> schrieb am Di., 7. Sept. 2021,
11:51:

R/draws-index.R
alyst: My version is here: https://github.com/alyst/posterior/tree/fix-188.
It's using 2 *match()* calls instead of *match()* and *merge()*, so potentially could be faster.
I haven't run any benchmarks, though.
Also it's possible that replacing *match()* with *vctrs::vec_match()* may speed it up, but I didn't test.

paul-buerkner:
thank you! would you mind testing it though? I cannot run the tests myself
at the moment.

Alexey Stukalov ***@***.***> schrieb am Di., 7. Sept. 2021,
12:56:

alyst:
> thank you! would you mind testing it though? I cannot run the tests myself at the moment

I can, but I'm not familiar with how it's usually done for this package. Are there some standard benchmarks in the source tree already? cc @mjskay

karldw:
@alyst, I was focusing on the timing of this test:
https://github.com/stan-dev/posterior/blob/0c0c50f35ae1c678aab5799e5b9fc8c65a85a056/tests/testthat/test-subset_draws.R#L116-L123

Though there are other cases that might matter, like the heavy loops you mentioned.

I've added you to my fork in case it makes things easier.

mjskay:
@alyst you should be able to run the full test suite with `devtools::test()`

alyst:
@mjskay The test suite passes on my laptop (including the performance test, I guess). But I thought there are some benchmarks where we can measure the performance more precisely.

mjskay:
Ah, I don't think it's quite so systematic yet unfortunately :). The plan is to use {touchstone} to do this eventually but it hasn't been done yet (see #159).

The test @karldw pointed to above is what I've been using informally to test it as well, via something like `bench::mark(subset_draws(x, colnames(x)), min_iterations = 100)`

alyst:
So with `bench::mark(subset_draws(x, colnames(x)), min_iterations = 100)`
for this PR (without my changes) I have:
```
r$> bench::mark(subset_draws(x, colnames(x)), min_iterations = 100)                                                                                                                                                  
# A tibble: 1 × 13
  expression                        min   median `itr/sec` mem_alloc `gc/sec` n_itr  n_gc total_time result                    memory time             gc                
  <bch:expr>                   <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl> <int> <dbl>   <bch:tm> <list>                    <list> <list>           <list>            
1 subset_draws(x, colnames(x))    491ms    554ms      1.65        NA     4.27   100   259      1.01m <drws_mtr [10 × 300,000]> <NULL> <bench_tm [100]> <tibble [100 × 3]>
Warning message:
Some expressions had a GC in every iteration; so filtering is disabled. 
```

With my changes:
```
# A tibble: 1 × 13
  expression                        min   median `itr/sec` mem_alloc `gc/sec` n_itr  n_gc total_time result                    memory time             gc                
  <bch:expr>                   <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl> <int> <dbl>   <bch:tm> <list>                    <list> <list>           <list>            
1 subset_draws(x, colnames(x))    154ms    166ms      5.59        NA     6.76   100   121      17.9s <drws_mtr [10 × 300,000]> <NULL> <bench_tm [100]> <tibble [100 × 3]>
Warning message:
Some expressions had a GC in every iteration; so filtering is disabled. 
```

Looks like it's 3x faster.

mjskay:
Awesome! Thanks for this :)

alyst:
You can just cherry-pick the commit from my branch into this PR.

mjskay:
Done. Thanks again!

@paul-buerkner if the checks pass now I think this is good to go.

paul-buerkner:
Thank you!

Issue #69, opened 2020-05-26
Hidden columns are confusing


bbbales2:
When I'm doing diagnostic stuff I like to be able to separate chains.

So them I'm writing with the hidden `.chain` and `.iteration` variables:

```
fit1$sampler_diagnostics() %>%
  as_draws_df() %>%
  ggplot(aes()) +
  geom_point(aes(.iteration, n_leapfrog__)) +
  facet_grid(.chain ~ .)
```

I guess now that I know they're there it's not a problem, but it's easier to know that all the columns in my dataframe are visible.

Why are these hidden?

paul-buerkner:
I thought it is better to only show actual variables by default, this this can be changed using the `meta_columns` argument of the print method. The question is whether this argument should be TRUE or FALSE by default and I am happy to discuss the pros and cons of these options.

bbbales2:
Is there any way to indicate the hidden columns print of the table? Like for example in the comments at the top of the print:

```
# A draws_df: 1000 iterations, 4 chains, and 9 variables (hidden columns .chains and .iterations)
```

I don't know if that's less confusing or likely to get people on the right track. I think I found out about these secret columns watching Jonah.

It does seem convenient to hide them. In what is presumably the most important use case (interpreting posterior samples) exposing `chain` and `iterations` would just mean everyone does a `select(-chain, -iterations)` as step-1, which doesn't seem ideal.

paul-buerkner:
That sounds like a good idea. I will add a note to the print output in that regard.

paul-buerkner:
How about:

```R
x <- as_draws_df(example_draws())
print(x)
# A draws_df: 100 iterations, 4 chains, and 10 variables
     mu tau theta[1] theta[2] theta[3] theta[4] theta[5] theta[6]
1  2.01 2.8     3.96    0.271    -0.74      2.1    0.923      1.7
2  1.46 7.0     0.12   -0.069     0.95      7.3   -0.062     11.3
3  5.81 9.7    21.25   14.931     1.83      1.4    0.531      7.2
4  6.85 4.8    14.70    8.586     2.67      4.4    4.758      8.1
5  1.81 2.8     5.96    1.156     3.11      2.0    0.769      4.7
6  3.84 4.1     5.76    9.909    -1.00      5.3    5.889     -1.7
7  5.47 4.0     4.03    4.151    10.15      6.6    3.741     -2.2
8  1.20 1.5    -0.28    1.846     0.47      4.3    1.467      3.3
[...]

PR #213, opened 2021-11-14
Tibble fixes

### LINKED ISSUE #202: Drop draws_df class if meta-data removed by dplyr verbs
From [this comment](https://github.com/stan-dev/posterior/issues/69#issuecomment-914659080):

> Reading more about this on the dplyr side, probably the right thing to do is for us to implement `dplyr::dplyr_reconstruct()` for the `"draws_df"` class in a similar way to `[.draws_df` so that it drops the class if required columns are removed (possibly with a warning like `[.draws_df` does, or maybe not?). We should be able to do this while keeping `dplyr` in suggests. See [here](https://dplyr.tidyverse.org/reference/dplyr_extending.html).

### LINKED ISSUE #210: implement pillar::format_glimpse() for rvars
Currently `pillar::glimpse()` fails if a data frame has an rvar in it:

```r
> library(posterior)
> data.frame(x = rvar(1:10))
              x
1 5.5 ± 3.02765
> pillar::glimpse(data.frame(x = rvar(1:10)))
Rows: 1
Columns: 1
Error: C stack usage  15922912 is too close to the limit
```

This can be fixed by implementing `pillar::format_glimpse()`, which should be pretty easy. Basically:

- [ ] if vector-like rvar, return character vector of elements converted to a display format. Some variant of `format.rvar()` should work
- [ ] if array-like (2 or more dims), return an abbreviation similar to what matrices / arrays do; e.g. `pillar::format_glimpse(matrix(1:10))` returns `"<matrix[10 x 1]>"`

mjskay:
#### Summary

This PR closes two minor tibble-related issues:

- closes #210 by implementing `pillar::format_glimpse()`. Pillar is already a dependency as it is needed for formating rvar output in tibbles, so this adds no dependencies. Implementing this function allows users of `pillar::glimpse()` to get nice output:

   ```r
   > library(posterior)
   > pillar::glimpse(tibble::tibble(mu = 1:10, x = rvar_rng(rnorm, 10, mu)))
   Rows: 10
   Columns: 2
   $ mu <int> 1, 2, 3, 4, 5, 6, 7, 8, 9, 10
   $ x  <rvar> 1 ± 0.97, 2 ± 0.99, 3 ± 1.00, 4 ± 0.99, 5 ± 1.00, 6 ± 1.01, 7 ± 1.00, 8 ± 1.00, 9 ± 1.00~
   ```

- closes #202 by implementing `dplyr::reconstruct()` to remove the `"draws_df"` and `"draws"` classes from a dataframe if {dplyr} verbs drop the metadata columns. For example, the following previously did not work (as reported by @ASKurz [here](https://github.com/stan-dev/posterior/issues/69#issuecomment-914508109)):

   ```r
   > example_draws() %>% as_draws_df() %>% count(.chain)
   # A tibble: 4 x 2
     .chain     n
      <int> <int>
   1      1   100
   2      2   100
   3      3   100
   4      4   100
   ```

   This is slightly trickier as {dplyr} is in Suggests, so to support R < 3.6 we have to register the S3 method in .onLoad rather than `@export`ing it.
   
   **A point of discussion**: unlike with `[.draws_df`, which produces a warning when the `"draws_df"` and `"draws"` classes are dropped, I opted not to do so here as it strikes me that any use of {dplyr} verbs that drop these columns are likely to be operations where the user no longer wants the functionality of the draws class on the result (e.g. summarization). This also makes me wonder if the warning is needed in `[.draws_df`? Happy to discuss alternatives here.

#### Copyright and Licensing


mjskay:
(the CI checks revealed a small problem with casting from rvars to distributional objects in R < 4, so I just fixed that on this PR as well.)

paul-buerkner (approved):
Thank you! Looks good to me. Waiting for checks to finish before I merge.

paul-buerkner:
It seems the remaining tests don't really want to run at moment. Since checks are passing locally for me too, I am merging this PR now.

mjskay:
Ah yup - that's because Ubuntu 16.04 images are obsolete on github ci now. We'll need to change to 18.04 (I just had to do this on another repo).

jgabry:
Cool, thanks for implementing these!

Issue #232, opened 2022-03-25
Better mcse_sd


avehtari:
The current mcse_sd assumes normality. More accurate mcse_sd for non-normally distributed variables can be obtained using moments of moments and first order Taylor series approximation as described by Kenney and Keeping (1951, p. 141) (Mathematics of Statistics, Part two, 2nd edition). I also made simulations to find out better ESS choice. The previous ess_sd was also ad-hoc and was a conservative choice that that underestimated ESS. New ess_mean(abs(sims-mean(sims))) is less ad-hoc, as it is more related to the actual computation for sd, and empirically it provides a good ESS with increasing S. 

The new computation has marginally very low bias, and the remaining excess conditional error is unavoidable due to the structure of the estimation problem: a) with small S, observed sd has high variation, b) when observed sd is smaller than asymptotic sd, MCSE which is based on sd is also smaller, and then standardized errors for small sd are big, and these dominate the error plots.

The following plots show the root mean square of the conditional standardized errors for marginally normal(0,1) and marginally Chi(1) distributed variables with underlying correlation from AR(1, phi) process. The same simulation were used in Rhat online postscript appendix https://avehtari.github.io/rhat_ess/ess_comparison.html

![image](https://user-images.githubusercontent.com/6705400/160079947-be8516b4-7e8b-47eb-b5ea-d6233fa8ad19.png)

avehtari:
closed by #233

PR #233, opened 2022-03-25
better mcse_sd

avehtari:
#### Summary

PR for issue #232 better mcse_sd

#### Copyright and Licensing

### DIFF
--- a/R/convergence.R
+++ b/R/convergence.R
@@ -332,7 +332,7 @@ ess_mean.rvar <- function(x, ...) {
 #' @template args-conv
 #' @template args-methods-dots
 #' @template return-conv
-#' @template ref-vehtari-rhat-2019
+#' @template ref-vehtari-rhat-2021
 #'
 #' @examples
 #' mu <- extract_variable_matrix(example_draws(), "mu")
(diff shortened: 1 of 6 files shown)


R/convergence.R
```diff
@@ -322,41 +321,6 @@ ess_mean.rvar <- function(x, ...) {
   summarise_rvar_by_element_with_chains(x, ess_mean, ...)
 }
 
-#' Effective sample size for the standard deviation
```
paul-buerkner, line 325: You seem to have removed the `ess_sd` function. Was that on purpose? We cannot just remove previously exported functions even if we don't use them anymore in the mcse computation. 

I suggest we add them back and only change the `mcse_sd` function.

paul-buerkner (changes requested):

R/convergence.R
avehtari: I put it back with the formula used in mcse_sd

paul-buerkner: Thank you!

paul-buerkner:
I will make some more manual checks and perhaps fix tests before merging but otherwise looks good.

paul-buerkner:
Merging now after a few minor changes.

paul-buerkner (approved):

Issue #246, opened 2022-07-08
diag() for rvars


mjskay:
Note to self (and from #245): make sure all uses of `diag()` are supported by rvar. Currently some are, but we will need to make it generic to support all uses.

PR #244, opened 2022-07-05
allow NULL rvars with ndraws() > 1, for #242

mjskay:
#### Summary

This is a draft PR that allows NULLish rvars to have more than one draw. Previously these were normalized to 1 draw of nothing; this caused incorrect draws_rvars objects to be created:

```r
> draws_rvars(x = rvar(NULL), y = rvar(rnorm(100)))
# A draws_rvars: 1 iterations, 1 chains, and 2 variables
$x: rvar<1>[0] mean ± sd:
NULL

$y: rvar<100>[1] mean ± sd:
[1] 0.14 ± 0.92 
```

With this PR, that is fixed:

```r
> draws_rvars(x = rvar(NULL), y = rvar(rnorm(100)))
# A draws_rvars: 100 iterations, 1 chains, and 2 variables
$x: rvar<100>[0] mean ± sd:
NULL

$y: rvar<100>[1] mean ± sd:
[1] 0.0022 ± 1.1 
```

Currently leaving this as a draft as the question remains if it makes sense to allow rvars with 0 draws, which may take a bit more thought to implement to handle corner cases. See discussion at #242.

#### Copyright and Licensing

By submitting this pull request, the copyright holder is agreeing to 
license the submitted work under the following licenses:

- Code: BSD 3-clause (https://opensource.org/licenses/BSD-3-Clause)
- Documentation: CC-BY 4.0 (https://creativecommons.org/licenses/by/4.0/)

Issue #243, opened 2022-07-05
as_draws_rvars() cannot parse indices with nested brackets


wds15:
I am dealing with stanfit objects which have posteriors of array for which I named the indices. This fails in some cases while it works in others:

```R
library(posterior)
loaded <- library(OncoBayes2, logical.return=T)
if(!loaded) {
    install.packages("OncoBayes2")
    library(OncoBayes2)
}

## run an OB2 example
example_model("single_agent")

## blrmfit$stanfit now contains a posterior with named arrays
names(blrmfit$stanfit)

## does work in this case:
post_rv <- as_draws_rvars(blrmfit$stanfit)
## looks great
subset_draws(post_rv, variable="beta_group")

## does not work:
example_model("combo2")
names(blrmfit$stanfit)

## throws an error:
post_rv <- as_draws_rvars(blrmfit$stanfit)
```

Maybe I am using the interface not as intended, but what makes me feel something is odd is that a simple example works (single-agent) while the more complex one fails.

Some more printouts:

stanfit which can be converted to rvars:
```R
> names(blrmfit$stanfit)
 [1] "log_beta_raw[1,1,1]"                           
 [2] "log_beta_raw[2,1,1]"                           
 [3] "log_beta_raw[1,1,2]"                           
 [4] "log_beta_raw[2,1,2]"                           
 [5] "mu_log_beta[log(drug_A/dref),intercept]"       
 [6] "mu_log_beta[log(drug_A/dref),log_slope]"       
 [7] "tau_log_beta_raw[1,1,1]"                       
 [8] "tau_log_beta_raw[1,1,2]"                       
 [9] "L_corr_log_beta[1,1,1]"                        
[10] "L_corr_log_beta[1,2,1]"                        
[11] "L_corr_log_beta[1,1,2]"                        
[12] "L_corr_log_beta[1,2,2]"                        
[13] "beta[1,1,1]"                                   
[14] "beta[2,1,1]"                                   
[15] "beta[1,1,2]"                                   
[16] "beta[2,1,2]"                                   
[17] "tau_log_beta[1,log(drug_A/dref),intercept]"    
[18] "tau_log_beta[1,log(drug_A/dref),log_slope]"    
[19] "beta_EX_prob[trial_A,log(drug_A/dref)]"        
[20] "beta_group[trial_A,log(drug_A/dref),intercept]"
[21] "beta_group[trial_A,log(drug_A/dref),slope]"    
[22] "log_lik_group[trial_A]"                        
[...]

Issue #245, opened 2022-07-08
structured rvars


wds15:
Hi!

rvars are a great thing and I have started using them a bit. I have run into issues when using it and there could be a nice solution but let me start with describing my setting. I wanted to transfer code from the `transformed parameters` block into R in order to eventually handle post-processing of my draws in R instead of Stan. Here are the respective Stan bits:

```stan
parameters {
  // the first 1:num_groups parameters are EX modelled while the
  // num_groups+1:2*num_groups are NEX
  vector[2] log_beta_raw[2*num_groups,num_comp];
  vector[num_inter] eta_raw[2*num_groups];

  // hierarchical priors
  vector[2] mu_log_beta[num_comp];
  // for differential discounting we allow the tau's to vary by
  // stratum (but not the means)
  vector<lower=0>[2] tau_log_beta_raw[num_strata,num_comp];
  cholesky_factor_corr[2] L_corr_log_beta[num_comp];

  vector[num_inter] mu_eta;
  vector<lower=0>[num_inter] tau_eta_raw[num_strata];
  cholesky_factor_corr[num_inter] L_corr_eta;
}
transformed parameters {
  vector[2] beta[2*num_groups,num_comp];
  vector[num_inter] eta[2*num_groups];
  vector<lower=0>[2] tau_log_beta[num_strata,num_comp];
  vector<lower=0>[num_inter] tau_eta[num_strata];

  // in the case of fixed tau's we fill them in here
  if (prior_tau_dist == 0) {
    tau_log_beta = prior_EX_tau_mean_comp;
    tau_eta = prior_EX_tau_mean_inter;
  } else {
    tau_log_beta = tau_log_beta_raw;
    tau_eta = tau_eta_raw;
  }

  // EX parameters which vary by stratum which is defined by the group
  for(g in 1:num_groups) {
    int s = group_stratum_cid[g];
    for(j in 1:num_comp) {
      beta[g,j] = mu_log_beta[j] +
        diag_pre_multiply(tau_log_beta[s,j], L_corr_log_beta[j]) * log_beta_raw[g,j];
    }
    if (num_inter > 0)
      eta[g] = mu_eta + diag_pre_multiply(tau_eta[s], L_corr_eta) * eta_raw[g];
  }

  // NEX parameters
  beta[num_groups+1:2*num_groups] = log_beta_raw[num_groups+1:2*num_groups];
  eta[num_groups+1:2*num_groups] = eta_raw[num_groups+1:2*num_groups];

[...]

PR #247, opened 2022-07-16
rvar indexing and casting improvements

### LINKED ISSUE #242: issue with 0-sized rvars
0-sized rvars behave unexpectedly as they look like 1-sized rvars and they even throw errors when converted to a matrix:

```R
mu <- as_rvar(rep(0, 0))
mu
as_draws_matrix(mu)
```

produces:

```R
> mu <- as_rvar(rep(0, 0))
> mu
rvar<1>[0] mean ± sd:
NULL
> as_draws_matrix(mu)
Error in dimnames(draws_of(x)) <- `*vtmp*` : 
  length of 'dimnames' [2] not equal to array extent
```

I would expect to get a 0 sized rvar (not a 1-sized as it looks) and when converting to a matrix I would expect to get a matrix with 0 columns and rows as many as there are samples (0 in this case).

mjskay:
#### Summary

This PR supersedes #244 and closes #242, #243, and #246. Specifically, it:

* Adds an implementation of `drop()` for `rvar`s.
* Supports the remaining modes of `diag()` for `rvar`s (#246).
* Uses more robust parsing for named indices in `as_draws_rvars()`, including nested use
of `[`, like `x[y[1],2]` (#243).
* Allows 0-length `rvar`s with `ndraws() > 1` (#242). e.g.
   ```r
   > rvar(array(numeric(), dim = c(10,0)))
   rvar<10>[0] mean ± sd:
   NULL
   ```
* Ensures 0-length `rvar`s can be cast to `draws` formats (#242); e.g. 
   ```r
   > as_draws_matrix(rvar(array(numeric(), dim = c(10,0))))
   # A draws_matrix: 10 iterations, 1 chains, and 0 variables
       variable
   draw
     1 
     2 
     3 
     4 
     5 
     6 
     7 
     8 
     9 
     10
   ```
* Includes some miscellaneous improvements to testing/coverage for rvars.

Also worth pointing out is that this PR overrides some base functions (`diag()` and `drop()`) using S4 instead of S3 (hence the added `methods` dep). This is because those functions are overloaded as S4 generics in other packages folks might be using (particularly `Matrix`), so I think this is the best way to prevent weird incompatibilities if both packages are loaded.

#### Copyright and Licensing


mjskay:
@wds15 I would be curious if you have feedback on this PR, as these issues were largely ones you raised. I believe the only outstanding thing to figure out is whether or not to make `drop = TRUE` the default (i.e. #245), but I think that will take a bit more thought and so can be a later PR.

wds15:
Great to see this pr. What do you suggest to test it? Should I recode the example I posted previously?

mjskay:
Sure! Yeah, hopefully the weird indices should parse properly now and you can at least do diag() without rdo(). I'm not sure in what context you were using 0-length rvars but hopefully that is fixed as well.

wds15:
Ok, will do, the 0 length thing is triggered by a unit test of the OncoBayes2 package. Let me report back.

mjskay:
This PR also fixes #248 now.

wds15:
I tested the things I reported:

- 0 sized rvars
- diag for rvars
- having "[]" in index names

all good as it looks. 

Just one thought:

image

It's not `NULL`, but rather 0 sized. Other than that the 0 sized `rvar` appears to behave fine.

I was not quite sure what to do about `drop`? The thing with drop is that I usually really want to control which index gets killed. This is why I often prefer `adrop` which allows me to control which dimension is being killed. So maybe add a `adrop` which works with `rvars`?

As I am going to rely with `OncoBayes2` in its next version on posterior, I am going to hold that release until these updates have made it to CRAN. So I am looking forward to see it in a 1.2.3 release.

Thanks a lot for addressing these things quickly. Much appreciated.

mjskay:
> I tested the things I reported:
> 
> * 0 sized rvars
> * diag for rvars
> * having "[]" in index names
> 
> all good as it looks.

Awesome, thanks for taking a look!

> image
> 
> It's not `NULL`, but rather 0 sized. Other than that the 0 sized `rvar` appears to behave fine.

Good point. Since the typical behavior on other zero-length objects seems to be to print the constructor that results in a zero-length object, I've changed the behavior to this:

```r
rvar()
## rvar<1>[0] mean ± sd:
## rvar()
```

> 
> I was not quite sure what to do about `drop`? The thing with drop is that I usually really want to control which index gets killed. This is why I often prefer `adrop` which allows me to control which dimension is being killed. So maybe add a `adrop` which works with `rvars`?

Fortunately, because `adrop()` appears to be implemented in terms of `dim()` and `dimnames()`, which already work on `rvar()`, `adrop()` appears to already work on `rvar()`! Since this is desirable behavior, I added a test or two to make sure it continues to work in the future.

> 
> As I am going to rely with `OncoBayes2` in its next version on posterior, I am going to hold that release until these updates have made it to CRAN. So I am looking forward to see it in a 1.2.3 release.

Cool!! Not sure when @paul-buerkner is planning a new release. Once this PR is in I plan to submit #236 shortly after that, which together should cover a bunch of outstanding `rvar()` things.

> Thanks a lot for addressing these things quickly. Much appreciated.

Thanks you for taking the time to test things and write it up! It's very exciting to see it being put through its paces :)

paul-buerkner:
I think we can schedule a new release flexibly. We have not many pressing issues that need to be resolved first (I think). If I schedule a release in the first week of Augst, would that work for this issue and for Sebastians package?

wds15:
Works for me. Thanks!

mjskay:
Works for me too. This PR is ready to be merged when you have a chance to look at it. After that I will submit #236 and I think that's all on my plate for the next release. 

There are some bigger things I want to get to for a future release (like #149 and #208) but they will take a bit more thought and time.

paul-buerkner (approved):
Looks good to me. Thank you!

Issue #239, opened 2022-05-17
Attribute or other way to distinguish MCMC vs MC


avehtari:
The posterior package started with focus on multi-chain MCMC and stores chain and iterations ids. These are useful when computing multi-chain Rhat, ESS, and MCSE. It is also possible to set weights for the draws which is usedul for importance sampling. It would be useful to think about the default behavior of some functions and whether the current draws objects contain sufficient information to do the right thing. For example, if we want to compute MCSE we have 4 different cases
- MCMC draws (e.g. usual Stan posterior draws, use MCMC-ESS to compute MCSE)
- MCMC draws with weights (e.g. Stan posterior draws + IS, e.g. in loo, use MCMC-ESS and IS-ESS)
- MC draws (e.g. draws from Gaussian posterior approximation, use MC-ESS)
- MC draws with weights (e.g. draws Gaussian posterior approximation + IS, use IS-ESS)

I guess we could assume that if iteration information is available, then the draws are from MCMC. But at the moment, we don't have support for independent (weighted) MC draws. Would it make sense to set the iteration to 1 for all independent draws? Other ideas for making the difference?

This issue is related to `psis()` function in loo package complaining if `r_eff` argument is not set. `r_eff` is used to pass the earlier computed (MCMC-ESS)/S. If we could determine whether the draws are from MCMC or MC, we would not need to complain in the latter case (and could compute r_eff internally in the first case)

paul-buerkner:
Not all formats have iteration information. In fact, in a way, only draws_df has. The other ones just store iteration implicitly through the ordering of the draws.

Adding an attribute would not be difficult I think. The only question is how to proceed with it when the draws objects are transformed somehow. Attributes in R are timid things that tend to vanish into the dark as soon as one lightly touches the object they belong to. 

For rvars it might be the easiest(?) to maintain an attribute as all transformations are fully custom there anywhere. For the other objects, I am not sure. I didn't want to go through the effort or reimplementing every standard transformation such as +, * etc. for every format to make sure the attributes are kept/alterted correctly.

Does anybody have other ideas how to differentiate this types of draws?

mjskay:
> For rvars it might be the easiest(?) to maintain an attribute as all transformations are fully custom there anywhere.

Yeah, the annoying stuff to make this work in rvars has already been figured out to track chain information, so it could certainly be done. It does seem like we wouldn't want these feature to be limited to rvars though.

[...]

Issue #149, opened 2021-05-21
discrete rvars (factors, ordered factors)


mjskay:
This is related to #17 but slightly more specific. At some point I think it would be good to have rvars that behave like factors, as at the very least these would be useful for posterior predictive distributions from ordinal and categorical models.

I think the general idea would be:
- have classes like `c("rvar_factor", "rvar")` and `c("rvar_ordered", "rvar_factor", "rvar")` with implementations of the base factor functions like `levels()`.
- have typical constructor and conversion functions like `rvar_factor()` and `as_rvar_factor()`
- (possibly) auto-detect factor types in `as_rvar()` by creating a factor if the input is an integer array with a `"levels"` attribute (as I think brms returns for posterior predictive distributions on categorical models).

This does raise a question if we similarly need subtypes like `"rvar_numeric"`, `"rvar_integer"`, and `"rvar_logical"` which could be automatically applied by the `rvar()` constructor, `as_rvar()`, and `draws_of()`. 

Another question is how to print them. For ordered factors maybe median +/- mad? For categorical factors I dunno (modal category with its proportion?)

I don't see any rush on this issue but I thought I'd throw it out there for thoughts/feedback.

---

Collecting implementation requirements here:
- [x] rvar_factor and rvar_ordered constructors and conversion functions
- [x] rvar / as_rvar should auto-detect factor-like input (arrays with `"levels"` attribute
- [x] rvar / as_rvar should auto-detect character input and convert to factor
- [x] figure out printing
- [x] figure out functions that should not work with factors
- [x] casting to/from other draws formats with mixed variables (some factor, some not)
- [x] implement `%in%` or `match`

jgabry:
> * (possibly) auto-detect factor types in `as_rvar()` by creating a factor if the input is an integer array with a `"levels"` attribute (as I think brms returns for posterior predictive distributions on categorical models).

That would be cool. There are also packages (e.g. rstanarm and at least one more that I'm forgetting at the moment) that return character for the ppd in these models, so maybe we could also convert character into factor automatically. 

> This does raise a question if we similarly need subtypes like `"rvar_numeric"`, `"rvar_integer"`, and `"rvar_logical"`

I can see the use case for factor more clearly, but I could easily be overlooking something. Would the distinction between numeric an integer have any real affect in practice? I'm thinking it only would if we used this information to prevent certain operations on discrete variables (but left to its own devices R would just treat numeric and integer the same in nearly every case, right?). 

[...]

Issue #17, opened 2019-10-29
How to handle discrete parameters?


MansMeg:
Im not sure if you have discussed this before? I guess it is important to handle since many non-stan models may have discrete parameters.

PR #258, opened 2022-10-09
Factor rvars

mjskay:
### Summary

This is a PR for adding factor-like `rvar`s, for #149.

Basic idea is that the backing array can be a `factor` or `ordered` with a `"levels"` attribute. These are auto-detected when objects with a `"levels"` attribute are passed to `rvar()`, or can be created explicitly with `rvar_factor()`, `rvar_ordered()`, `as_rvar_factor()`, or `as_rvar_ordered()`.

A few other changes are in here, summarized in NEWS.md.

### Examples

```r
x = rvar(array(c("a","a","b","c","d","d","d","a"), dim = c(4,2)))
x
#> rvar_factor<4>[2] mode <entropy>:
#> [1] a <0.75>  d <0.41> 
#> 4 levels: a b c d
```

Currently the summary output in vector / array layouts is modal category (`modal_category()`) with **normalized** entropy (`entropy()`, which is normalized to be between 0 (all one category) and 1 (max entropy: uniform)).

Ordered factors are similar, but show `mode <dissent>`:

```r
x = rvar_ordered(array(c("a","a","b","c","d","d","d","a"), dim = c(4,2)))
x
#> rvar_ordered<4>[2] mode <dissent>:
#> [1] a <0.43>  d <0.81> 
#> 4 levels: a < b < c < d
```

Where `dissent()` is [Tastle and Wierman](https://doi.org/10.1016/j.ijar.2006.06.024)'s "dissention" measure (thanks @isabellaghement), and ranges from 0 (all in one category) to 1 (bimodal at opposite ends of the scale).

Here's another example on output of `posterior_predict()`, which should work for both `rstanarm::stan_polr()` (which uses strings) and `brms()` (which uses an array with a `"levels"` attribute):

```r
library(brms)

df <- esoph
m <- brm(tobgp ~ agegp, data = df, family = cumulative, seed = 12345, backend = "cmdstanr")

df$y_rep = rvar_ordered(posterior_predict(m))
head(df, n = 15)
#>    agegp     alcgp    tobgp ncases ncontrols           y_rep
#> 1  25-34 0-39g/day 0-9g/day      0        40    10-19 <0.63>
#> 2  25-34 0-39g/day    10-19      0        10    10-19 <0.63>
#> 3  25-34 0-39g/day    20-29      0         6    10-19 <0.62>
#> 4  25-34 0-39g/day      30+      0         5 0-9g/day <0.64>
#> 5  25-34     40-79 0-9g/day      0        27    10-19 <0.63>
#> 6  25-34     40-79    10-19      0         7 0-9g/day <0.64>
#> 7  25-34     40-79    20-29      0         4    10-19 <0.62>
#> 8  25-34     40-79      30+      0         7    10-19 <0.63>
#> 9  25-34    80-119 0-9g/day      0         2    10-19 <0.64>
#> 10 25-34    80-119    10-19      0         1 0-9g/day <0.64>
#> 11 25-34    80-119      30+      0         2    10-19 <0.63>
#> 12 25-34      120+ 0-9g/day      0         1    10-19 <0.64>
#> 13 25-34      120+    10-19      1         0    10-19 <0.64>
#> 14 25-34      120+    20-29      0         1    10-19 <0.62>
#> 15 25-34      120+      30+      0         2    10-19 <0.62>
```

This PR also provides an implementation of `match()` and `%in%` for `rvar` based on a conversation with @bwiernik, which is helpful especially for `rvar_factor`; e.g.:

```r
x
#> rvar_ordered<4>[2] mode <dissent>:
#> [1] a <0.43>  d <0.81> 
#> 4 levels: a < b < c < d
x %in% c("a", "b")
#> rvar<4>[2] mean ± sd:
#> [1] 0.75 ± 0.5  0.25 ± 0.5
```

(this required adding generics for `match()` and `%in%`, but I think that is reasonable).

### Support in other formats

Currently, `draws_rvars`, `draws_list`, and `draws_df` can all support discrete variables (when converting to the list and df formats, factors and ordereds are used). The `draws_matrix` and `draws_array` formats do not support discrete variables. I thought about @paul-buerkner's suggestion to allow them to support such formats when all variables are of the same type, but concluded this would be pretty hacky: they would have to be either character arrays (which lose information about the order of levels, so could only be used for factors --- and even then, lose information about levels with 0 realizations) or would have to be factor-like arrays, which have poor support in base R (most operations on factor-like arrays just convert them to vectors). And even then, it would only work in situations where the levels of all variables are the same. The only way I can think of to support them comprehensively in those formats would be to add a parallel data structure as an attribute that stores type information for each column (is it a factor, is it ordered, what levels does it have, ...). I think we may end up having to do that anyway (e.g. to support #239), so maybe we'll get there eventually, but I think this PR is already pretty big.

So, ultimately, I decided to stick with lossy conversion for now: when converting a `draws_rvars`, `draws_list`, or `draws_df` to a `draws_array` or `draws_matrix`, if the source object contains factors or ordereds, a warning is given and the variable is converted to numeric.

For example:

```r
d = draws_rvars(x = rvar(c("a","a","b","c")))
d
#> # A draws_rvars: 4 iterations, 1 chains, and 1 variables
#> $x: rvar_factor<4>[1] mode <entropy>:
#> [1] a <0.95> 
#> 3 levels: a b c
as_draws_df(d)
#> # A draws_df: 4 iterations, 1 chains, and 1 variables
#>   x
#> 1 a
#> 2 a
#> 3 b
#> 4 c
#> # ... hidden reserved variables {'.chain', '.iteration', '.draw'}
as_draws_array(d)
#> Warning: draws_array does not support non-numeric variables (e.g., factors).
#> Converting non-numeric variables to numeric.
#> # A draws_array: 4 iterations, 1 chains, and 1 variables
#> , , variable = x
#> 
#>          chain
#> iteration 1
#>         1 1
#>         2 1
#>         3 2
#>         4 3
```

One current issue created by the fact that draws_array does not support discrete variables is that this conversion warning will be generated whenever discrete variables are used with `summarise_draws()`, as all formats currently convert to draws_array in `summarise_draws()`:

```r
summarise_draws(d)
#> # A tibble: 1 × 10
#>   variable  mean median    sd   mad    q5   q95  rhat ess_bulk ess_tail
#>   <chr>    <dbl>  <dbl> <dbl> <dbl> <dbl> <dbl> <dbl>    <dbl>    <dbl>
#> 1 x         1.75    1.5 0.957 0.741     1  2.85  2.56       NA       NA
#> Warning message:
#> draws_array does not support non-numeric variables (e.g., factors). Converting non-numeric variables to numeric. 
```

 I think that for now at least the warning is helpful to indicate that the output may be problematic. Long term, I'm not sure I see a way around this without either re-writing internals of `summarise_draws()`, adding variable-level type information to draws_array, or providing an alternative function, since currently `draws_array` fundamentally can't support heterogeneous variable types.

### Notes / Feedback

There are still several places feedback would be helpful:

- If anyone using categorical models can give this a test drive that'd be useful.
- Is the current solution to conversion to/from other draws formats acceptable? My thinking is, if we want more comprehensive support in all formats (i.e. array and matrix), that could be a future change related to #17 and #239.
- Is the current solution for `summarise_draws()` (warning on discrete variables) okay?
- Thoughts on summary functions for discrete parameters? Currently there is `modal_category()`, `entropy()` (which is **normalized** entropy; not sure if the name should be `normalized_entropy()` instead though that is rather wordy), and `dissent()`. I don't want to add a whole bunch, but would like the defaults for printing to be sensible in some way.
- Currently these things probably won't be that easy to visualize. {ggdist} will get some kind of support for them but I haven't figured that out yet.

### Copyright and Licensing


paul-buerkner:
Thank you for working on this! I don't have much input right now but at least a few thoughts.

- I like the `mode <entropy>` printing.
- Do we want to support categorical variables more generally in all formats? If yes, it could make sense to design a consistent interface across all at this point. Said differently, could there be any problems we run into later in other formats if we only implement categorical variables in rvars now?

mjskay:
> I like the `mode <entropy>` printing.

works for me

> Do we want to support categorical variables more generally in all formats? If yes, it could make sense to design a consistent interface across all at this point. Said differently, could there be any problems we run into later in other formats if we only implement categorical variables in rvars now?

Thinking about the different formats:

- `draws_list` and `draws_df` should have no problems with categorical variables, as they could just use `factor` and `ordered` types for those variables
- `draws_matrix` and `draws_array` I'm not sure what to do with, as every variable in an array/matrix must have the same type. One option is that we make converting to them be lossy by changing factors to integers. Another option would be to store a list of levels for each variable as an attribute of the draws object (e.g. similar to the way that dimnames are stored). But we would also need to store if they are ordered or not, and possibly handle modifications of the variables too... ultimately it might amount to reimplementing the factor type and not be worth it.

I think in the short term we could add support for factors to `draws_list` and `draws_df` easily, and go with the lossy-cast-to-integer approach for the other two formats. Then we can decide later if we want to deal with the complexity of adding support to them. I think this should be future-proof...?

isabellaghement:
Matthew, we chatted on Twitter about possible ways to summarize information on posterior distributions of categorical variables.  In particular, I wondered if the entropy reported for the posterior distribution of a categorical variable was computed differently depending on whether the categorical variable was nominal or ordinal.  (It didn't seem like it was.) 

In doing some more digging about this, I found an R package called **agrmt** [(agrmt: Calculate Concentration and Dispersion in Ordered Rating Scales)](https://cran.r-project.org/web/packages/agrmt/index.html) which purports to do the following: 

> Calculates concentration and dispersion in ordered rating scales. It implements various measures of concentration and dispersion to describe what researchers variably call agreement, concentration, consensus, dispersion, or polarization among respondents in ordered data. It also implements other related measures to classify distributions. In addition to a generic city-block based concentration measure and a generic dispersion measure, the package implements various measures, including van der Eijk's (2001) <[doi:10.1023/A:1010374114305](https://doi.org/10.1023%2FA%3A1010374114305)> measure of agreement A, measures of concentration by Leik, Tatsle and Wierman, Blair and Lacy, Kvalseth, Berry and Mielke, Reardon, and Garcia-Montalvo and Reynal-Querol. Furthermore, the package provides an implementation of Galtungs AJUS-system to classify distributions, as well as a function to identify the position of multiple modes.

On the surface of it, it sounds like this package should give you some more options in terms of describing posterior distributions of ordinal categorical variables. 

Here's some R code I put together to get you started - it is based on the package vignette available at [https://cran.r-project.org/web/packages/agrmt/vignettes/agrmt.pdf](https://cran.r-project.org/web/packages/agrmt/vignettes/agrmt.pdf) and on the help files available for this package.  In the code, I used a frequency vector mentioned at the beginning of the vignette, frequencies <- c(10, 20, 30, 15, 4), which corresponds to the following 5-category bar plot: 

image



`install.packages("agrmt")

library(agrmt)

categories <- c(1, 2, 3, 4, 5)
frequencies <- c(10, 20, 30, 15, 4)

freqtable <- data.frame(categories, frequencies)

freqtable$categories <- factor(freqtable$categories, ordered = TRUE)

library(ggplot2) 
library(magrittr)
theme_set(theme_bw())

plot_freqtable <- 
freqtable %>% 
  ggplot() + 
  aes(x = categories, y = frequencies, fill = categories) + 
  geom_col() +
  labs(x = "Category", y = "Frequency") + 
  theme(legend.position="none") 

plot_freqtable

# IMPORTANT NOTE: 
#
# Some functions in the agrmt package require frequency vectors while others require 
# standardized frequency vectors, that is the frequencies expressed as a proportion. 
# 
# Where standardized frequency vectors are required, frequency
# vectors that do not sum to 1 (i.e. that are not standardized) are automatically
# standardized by dividing each element of the frequency vector by the sum of
# the vector. This assumes complete data; a general assumption of the measures
# in this package.

#==============================================================
# 1) Agreement: Levels of agreement range from −1 to 1
# 
# Note: Agreement is only defined if there are at least three response categories, 
# and it does not tell you which of the categories is the most common one. 
# For this reason, it is also advisable to look at an appropriate measure of 
# central tendency, such as the interpolated median.
#==============================================================

# To calculate agreement, we simply call agreement with the frequency vector as
# the argument. There is normally no reason to use the old algorithm, which can
# be set using old=TRUE. This is the ‘old’ algorithm as described by Van der Eijk.

help(agreement, package = "agrmt") 

library(dplyr)

agreement_freqtable <- 
freqtable %>% 
  pull(frequencies) %>% 
  agreement() %>% 
  round(2)

agreement_freqtable  # 0.39

plot_freqtable + 
  ggtitle(paste("Agreement:", agreement_freqtable))
  

#==============================================================
# 2) Polarization
#
# The agreement scores are suitable to express polarization, but the numbers are
# not intuitive to interpret. The function polarization offers easily interpretable
# values by re-scaling agreement A. More precisely, it gives (1 − agreement)/2.
# This means that a polarization value of 1 means perfect polarization, 
# and a value of 0 means perfect agreement. A value of 0.5 corresponds to  
# ‘no agreement’.
#==============================================================

polarization_freqtable <- 
  freqtable %>% 
  pull(frequencies) %>% 
  polarization() %>% 
  round(2)

polarization_freqtable  # 0.30

#==============================================================
# 3) Leik’s Consensus (measure of ordinal dispersion):
# 
# If all observations are in the same category, ordinal dispersion is 0. 
# With half the observations in one extreme category, and half the observations 
# in the other extreme, Leik’s measure gives a value of 1.
#
# Leik’s ordinal dispersion measure is a percentage, and can be interpreted
# accordingly. [Important Note: R reports it as a proportion, not as a percentage!]
# 
# Ordinal dispersion can be used to express consensus or agreement, 
# simply by taking: 
#    1 - ordinal dispersion.
#==============================================================

help(Leik, package = "agrmt")

leik_freqtable <- 
  freqtable %>% 
  pull(frequencies) %>% 
  Leik() %>% 
  round(2)

leik_freqtable  # 0.40

#==============================================================
# 4) Tatsle and Wierman’s Consensus (computed using the Shannon entropy as the basis)
# 
#  It is 1 if there is perfect uniformity; 
#  it is 0 if there is perfect bimodality (=lack of agreement)
#==============================================================

help(consensus, package = "agrmt")

consensus_freqtable <- 
  freqtable %>% 
  pull(frequencies) %>% 
  consensus() %>% 
  round(2)

consensus_freqtable  # 0.62

#==============================================================
# 5) Entropy
# 
# The Shannon entropy can be calculated using the entropy() function. 
# This function follows TW (i.e., Tatsle and Wierman) and ignores categories 
# with zero observations. 
#
# The Shannon entropy ignores the order of categories; use the consensus() 
# function to consider the order of categories.
#==============================================================

help(entropy, package = "agrmt")

entropy_freqtable <- 
  freqtable %>% 
  pull(frequencies) %>% 
  entropy() %>% 
  round(2)

entropy_freqtable  # 2.08


#==============================================================
# Aside:   We can also classify a distribution using the AJUS-system introduced by Galtung (1969), 
# with the aim of reducing complexity. 
# 
# Distributions are classified as A if they are unimodal with a peak in the centre, 
# as J if they are unimodal with a peak at either end, 
# as U if they are bimodal with a peak at both ends, and as S if they are multimodal. 
# 
# In addition to Galtung's classification, the function classifies distributions 
# as F if there is no peak and all values are more or less the same (flat). 
# 
# Furthermore, a distinction is drawn between J and L distributions, 
# depending on whether they increase or decrease: J types have a peak on the right, 
# L types have the peak on the left. The skew is given as +1 for a positive skew, 
# as 0 for no skew, and -1 for a negative skew.
# 
# The skew is identified by comparing the sum of values left and right of the midpoint 
# respectively. For J-type of distributions, the skew is identified on the basis of the 
# changes between values. This way, long tails cannot influence the skew, 
# and a single peak at the left and right-hand end can be differentiated in all cases.
# 
#==============================================================

help(ajus, package = "agrmt")

freqtable %>% 
  pull(frequencies) %>% 
  ajus() 

`

My thinking is that the user could be given a menu of possibilities for describing the posterior distribution corresponding to an ordinal categorical variable via some of the functions in the **agrmt** package:  agreement(), polarization(), Leik(), consensus(), entropy(), etc.  (Note that the package includes other functions - I haven't explored them all.) 

Thanks, Matthew - I hope some of the above will be helpful to you.

Isabella

paul-buerkner:
Thank you @isabellaghement for checking out all these options!

mjskay:
Thanks @isabellaghement, this is very useful!

I think we can avoid a dep on {agrmt} at least for the default printing since the implementations of these are very straightforward, and I don't think we need a bunch of different measures of dispersion just for the rvar printing function (for continuous dists we only provide SD and MAD, for example). For `print.rvar()` I'd rather choose one or two reasonable defaults, then if people want more dispersion measures they can calculate them using `summarise_draws()`, not by setting the `summary` argument of `print.rvar()` (which is always awkward to use anyway, since people rarely call `print()` directly). 

After reading through these examples and looking at [Tastle and Wierman](https://doi.org/10.1016/j.ijar.2006.06.024), their "dissention" measure seems pretty reasonable to me: it's based on entropy, and ranges from 0 (all probability in one category) through 0.5 (uniform) to 1 (all probability split evenly between first and last category). So we could use that as a default for ordinal factors, entropy for categorical, and if people want something else they can calculate it with `summarise_draws()`.

mjskay:
Thinking about summary functions raises another question for @paul-buerkner: do we want the `entropy()` and `dissent()` functions to be exported so that people can use them in `summarise_draws()`? I suppose we may also want to export a version of `.mode()` (perhaps one that errors if applied to numeric data, and with a different name... something like `modal_category()` or `modal_level()` but maybe less wordy).

mjskay:
FYI this now has pretty comprehensive tests. All that remains is for me to update the `rvar` vignette and to make any adjustments based on folks' feedback.

Re: conversions to other formats, the current code supports conversion to/from `draws_df` and `draws_list` and will keep discrete variables as `factor`s or `ordered`s in those formats. When converting to `draws_matrix` or `draws_array`, the conversion is "lossy" and will result in numeric variables. Not sure if there's a better solution than that for now. Could raise a warning in those cases if desired?

paul-buerkner:
I think exporting these functions sounds good. I don't mind wordy names, so I am happy with both `modal_category` and `modal_level`, perhaps with a slight preference for the former. What would you prefer?

With regard to conversion, it should be possible to support *categorical only* `draws_array` and `draws_matrix`, right? Only if we mixed categorical and numerical variables, the conversion would be lossy (and then trigger a warning).

mjskay:
> I think exporting these functions sounds good. I don't mind wordy names, so I am happy with both modal_category and modal_level, perhaps with a slight preference for the former. What would you prefer?

`modal_category` works. TODO for me:
- [ ] export and document `modal_category`, `entropy`, `dissent`

> With regard to conversion, it should be possible to support categorical only draws_array and draws_matrix, right? Only if we mixed categorical and numerical variables, the conversion would be lossy (and then trigger a warning).

Hmm, fair point. It could get a little hairy since we'd probably have to apply a class (or at least an attribute) to remember that the arrays are factors or ordereds... and base-R `factor()` and `ordered()` typically do not work well as arrays (they tend to lose their dimensions). I can think about it and/or try something out to see what might work.

mjskay:
> > With regard to conversion, it should be possible to support categorical only draws_array and draws_matrix, right? Only if we mixed categorical and numerical variables, the conversion would be lossy (and then trigger a warning).
> 
> Hmm, fair point. It could get a little hairy since we'd probably have to apply a class (or at least an attribute) to remember that the arrays are factors or ordereds... and base-R `factor()` and `ordered()` typically do not work well as arrays (they tend to lose their dimensions). I can think about it and/or try something out to see what might work.

I've come to the conclusion that this isn't really worth it (I added rationale to the top comment). For now, I have gone with lossy conversion + a warning. Long term, we may want a way of storing type information for each variable in a draws_array/matrix, which would allow us to address this issue.

I believe this PR has now reached a "ready" state. I have updated the [summary comment](https://github.com/stan-dev/posterior/pull/258#issue-1402153039) to be more comprehensive, as well as NEWS.md. Feedback welcome!

mjskay:
The [factor_rvar](https://github.com/mjskay/ggdist/tree/factor_rvar) branch of {ggdist} now has experimental support for rvar_factors as well. Simple example of brms + posterior + ggdist with an ordinal model:

```r
library(brms)
library(posterior)
library(ggdist)

df <- esoph
m <- brm(tobgp ~ agegp, data = df, family = cumulative, seed = 12345, backend = "cmdstanr")

grid <- unique(df[,"agegp", drop = FALSE])
grid$pred <- rvar_ordered(posterior_predict(m, newdata = grid))

grid |>
  ggplot(aes(x = agegp, ydist = pred)) +
  stat_slab()
```
image

Relevant issue over there is: https://github.com/mjskay/ggdist/issues/108

R/as_draws_rvars.R
```diff
@@ -164,11 +169,24 @@ as_draws_rvars.draws_array <- function(x, ...) {
 
 #' @rdname draws_rvars
 #' @export
-as_draws_rvars.draws_df <- as_draws_rvars.draws_array
+as_draws_rvars.draws_df <- function(x, ...) {
+  x_df <- as.data.frame(x, optional = TRUE)[, variables(x, reserved = TRUE), drop = FALSE]
+  data_frame_to_matrix <- function(df) {
+    if (any(vapply(df, is.factor, logical(1)))) {
```
paul-buerkner, line 175: shall we have a separate function/method that does the check whether any of the variables in non-numeric in a draws object?

R/as_draws_rvars.R
```diff
@@ -164,11 +169,24 @@ as_draws_rvars.draws_array <- function(x, ...) {
...
+      # as.matrix() does not convert factor columns correctly, must do this ourselves
+      while_preserving_dims(function(df) do.call(function(...) vctrs::vec_c(..., .name_spec = rlang::zap()), df), df)
```
paul-buerkner, line 177: consider adding a line break in this and similarily very long lines?

R/discrete-summaries.R
```diff
@@ -0,0 +1,227 @@
...
+  UseMethod("entropy")
+}
+#' @rdname entropy
+#' @export
+entropy.default <- function(x) {
+  if (anyNA(x)) return(NA_real_)
+  p <- prop.table(simple_table(x)$count)
+  n <- length(p)
+  if (n == 1) {
+    0
```
paul-buerkner, line 64: I personally feel more comfortable with a single return variable at the end to avoid potential problems when extending the function. Something like 

```R
if (n == 1) {
  out <- 0
} else {
  out <- x
}
out 
# end of function
```

but I am also okay with the current pattern.

R/rvar-.R
```diff
@@ -322,6 +328,84 @@ all.equal.rvar <- function(target, current, ...) {
...
+#' When `x` is a base vector, a vector of the same length as `x`.
+#'
+#' When `x` is an [rvar], an [rvar] the same shape as `x`.
+#' @examples
+#' x <- rvar(c("a","b","b","c","d"))
+#' x %in% c("b","d")
+#'
+#' # for additional examples, see base::match()
+#' @export
+match <- function(x, table, ...) UseMethod("match")
```
paul-buerkner, line 365: perhaps add an empty line between end of function definitions and start of new funtion doc?

R/rvar-.R
```diff
@@ -585,6 +697,21 @@ flatten_array = function(x, x_name = NULL) {
   x
 }
 
+#' Fast conversion of rvar draws to a flattened data frame. Equivalent to
+#' as.data.frame(draws_of(flatten_array(x, name))) except it works with
+#' factor rvars (as.data.frame does not work on array-like factors)
+#' @noRd
+flatten_rvar_draws_to_df <- function(x, x_name = NULL) {
+  if (length(x) == 0) {
+    data.frame(row.names = draw_ids(x))
```
paul-buerkner, line 706: same here with regard to a single returned variable at the end.

R/rvar-.R
```diff
@@ -663,9 +791,62 @@ cleanup_draw_dims <- function(x) {
     rownames(x) <- as.character(seq_rows(x))
   }
 
+  # if x is factor-like (character or with "levels" attr), make it a factor
+  if (is.character(x)) {
```
paul-buerkner, line 795: is it worth making this "as.factor" code its own function for easy reuse and extention?

R/rvar-bind.R
```diff
@@ -69,7 +73,13 @@ bind_rvars <- function(args, arg_exprs, deparse.level = 1, axis = 1) {
 
 #' broadcast two rvars to compatible dimensions and bind along the `axis` dimension
 #' @noRd
-broadcast_and_bind_rvars <- function(x, y, axis = 1) {
+broadcast_and_bind_rvars <- function(x, y, axis = 1) UseMethod("broadcast_and_bind_rvars")
```
paul-buerkner, line 76: add a line break before `UseMethod?`

R/rvar-bind.R
```diff
@@ -69,7 +73,13 @@ bind_rvars <- function(args, arg_exprs, deparse.level = 1, axis = 1) {
...
+
+#' @export
+broadcast_and_bind_rvars.rvar <- function(x, y, axis = 1) {
+  # TODO: should really replace this with something that takes a list of rvars,
```
paul-buerkner, line 80: How urgent/important is that TODO? Is it worth including this already in the current PR?

R/rvar-bind.R
```diff
@@ -91,10 +101,21 @@ broadcast_and_bind_rvars <- function(x, y, axis = 1) {
...
+
   # bind along desired axis
-  result <- new_rvar(abind(draws_x, draws_y, along = draws_axis), .nchains = nchains(x))
+  result <- new_rvar(abind(draws_x, draws_y, along = draws_axis, use.dnns = TRUE), .nchains = nchains(x))
 }
 
+#' @export
+broadcast_and_bind_rvars.rvar_factor <- function(x, y, axis = 1) {
+  result <- NextMethod()
+  combine_rvar_factor_levels(result, list(levels(x), levels(y)), is_rvar_ordered(x) && is_rvar_ordered(y))
```
paul-buerkner, line 115: consider adding a line break.

R/rvar-cast.R
```diff
@@ -106,7 +126,23 @@ as.vector.rvar <- function(x, mode = "any") {
 
 #' @export
 as.list.rvar <- function(x, ...) {
-  apply(draws_of(x), 2, new_rvar, .nchains = nchains(x))
+  x_dim <- dim(x)
+  if (length(x_dim) >= 2) {
```
paul-buerkner, line 130: same here about a single returned variable at the end

R/rvar-cast.R
```diff
@@ -135,118 +171,300 @@ invalidate_rvar_cache = function(x) {
...
 #' @export
 vec_ptype2.rvar.double <- function(x, y, ...) new_rvar()
-
 #' @export
 vec_cast.rvar.double <- function(x, to, ...) new_constant_rvar(x)
 
+# double -> rvar_factor
 #' @export
-vec_ptype2.integer.rvar <- function(x, y, ...) new_rvar()
+vec_cast.rvar_factor.double <- function(x, to, ...) new_constant_rvar(while_preserving_dims(as.factor, x))
```
paul-buerkner, line 322: I suggest to add a line break before function body even in one-line functions

R/rvar-factor.R
```diff
@@ -0,0 +1,224 @@
...
+#' See [rvar()] for more details on the internals of the random variable datatype.
+#'
+#' @seealso [as_rvar_factor()] to convert objects to `rvar_factor`s. See [rdo()], [rfun()], and
+#' [rvar_rng()] for higher-level interfaces for creating `rvar`s.
+#'
+#' @return An object of class `"rvar_factor"` representing a `factor`-like random variable.
+#'
+#' @examples
+#'
+#' # TODO: update examples
```
paul-buerkner, line 29: Should this be TODO added to the PR before merging?

R/rvar-factor.R
```diff
@@ -0,0 +1,224 @@
...
+#' # If the input to rvar() is an array with a `"levels"` attribute, it
+#' # will automatically be treated as an `rvar_factor()`:
+#' y_array <- t(array(rbinom(3000, 1, c(0.1, 0.5, 0.9)) + 1, dim = c(3, 1000)))
+#' rvar(y_array)
+#' # with levels
+#' attr(y_array, "levels") = c("a", "b")
+#' rvar(y_array)
+#'
+#' @export
+rvar_factor <- function(x = factor(), dim = NULL, dimnames = NULL, nchains = NULL, with_chains = FALSE, ...) {
```
paul-buerkner, line 62: consider adding more line breaks here and below

R/rvar-math.R
```diff
@@ -26,7 +30,62 @@ Ops.rvar <- function(e1, e2) {
...
+  draws <- f(draws_x, draws_y)
+
+  # for factors, we must copy the dims back over after, because base-R factor
+  # operations destroy dims (while base-R numeric array operations do not)
+  if (preserve_dims) {
+    # in the case where one of e1 or e2 is scalar and the other is not,
+    # we need to source the dims from whichever is *not* scalar, which will be
+    # the longer of the two. In any other case, we give preference to e1
+    # (draws_x) to be consistent with base R behavior
+    dim_source <- if (length(draws_y) > length(draws_x)) {
```
paul-buerkner, line 42: this kind of assignment pattern is a bit harder to read than an explicit

```R
if () {
  dim_source = ...
} else {
  dim_source = ...
}
```

R/rvar-summaries-over-draws.R
```diff
@@ -210,25 +215,25 @@ range.rvar <- function(x, ...) {
 #' @rdname rvar-summaries-over-draws
 #' @export
 is.finite.rvar <- function(x) {
-  summarise_rvar_by_element_via_matrix(x, function(x) matrixStats::colAlls(is.finite(x), useNames = FALSE))
+  summarise_rvar_by_element_via_matrix(x, "is.finite", function(x) matrixStats::colAlls(is.finite(x), useNames = FALSE), .factor_okay = TRUE)
```
paul-buerkner, line 218: consider adding more line breaks to the very long lines here

R/rvar-summaries-within-draws.R
```diff
@@ -122,9 +126,13 @@ rvar_range <- function(..., na.rm = FALSE) {
 rvar_quantile <- function(..., probs, names = FALSE, na.rm = FALSE) {
   names <- as_one_logical(names)
   na.rm <- as_one_logical(na.rm)
+  x <- c(...)
+
+  type <- if (is_rvar_ordered(x)) 1 else 7
```
paul-buerkner, line 131: add a comment on what type 1 and 7 are and why there are chosen here?

tests/testthat/test-rvar-apply.R
```diff
@@ -42,6 +58,31 @@ test_that("rvar_apply works", {
...
   ref["a1",,] <- ref["a1",,] + 1
   ref["a2",,] <- 0
-  rvar_apply(x, 1, function(x) if(mean(draws_of(x)) > 18) rvar(0) else x + 1)
+  expect_equal(rvar_apply(x, 1, function(x) if(mean(draws_of(x)) > 18) rvar(0) else x + 1), ref)
 })
 
+
+# rvar_apply with factors -------------------------------------------------
+
+test_that("rvar_apply works with rvar_factor", {
```
paul-buerkner, line 67: Is there reason to believe this could work differently with `rvar_ordered`? In other words, would it make sense to test `rvar_ordered` variables here too?

tests/testthat/test-rvar-bind.R
```diff
@@ -18,13 +21,92 @@ test_that("c works on rvar", {
...
+  expect_equal(c(x, x), rvar(array(letters[c(1:9, 1:9)], dim = c(3,6))))
+
+  expect_equal(c(x, "a"), rvar(array(c(letters[1:9], "a", "a", "a"), dim = c(3,4))))
+  expect_equal(vctrs::vec_c("a", x), rvar(array(c("a", "a", "a", letters[1:9]), dim = c(3,4))))
+  expect_equal(c(x, factor("a")), rvar(array(c(letters[1:9], "a", "a", "a"), dim = c(3,4))))
+  expect_equal(vctrs::vec_c(factor("a"), x), rvar(array(c("a", "a", "a", letters[1:9]), dim = c(3,4))))
+  expect_equal(c(x, "xx"), rvar(array(c(letters[1:9], "xx", "xx", "xx"), dim = c(3,4))))
+  expect_equal(c(x, factor("xx")), rvar(array(c(letters[1:9], "xx", "xx", "xx"), dim = c(3,4))))
+  expect_equal(c(x, ordered("xx")), rvar(array(c(letters[1:9], "xx", "xx", "xx"), dim = c(3,4))))
+  expect_equal(vctrs::vec_c(ordered("xx"), x), rvar_factor(array(c("xx", "xx", "xx", letters[1:9]), dim = c(3,4)), levels = c("xx", letters[1:9])))
```
paul-buerkner, line 55: consider adding some more line breaks here and below

DESCRIPTION
```diff
@@ -31,7 +31,7 @@ Imports:
   rlang (>= 0.4.7),
```
paul-buerkner, line 31: Bump version number

paul-buerkner (changes requested):
This PR looks already really good, up to a few minor things, mostly style related.

I am impressed by the level of thought (and effort) you put into getting factors and ordered factors to work as natively as possible!

Some of the rvar internals remain a bit of black magic to me, so I don't understand every detail of it, but I trust you and the tests you put in place on that one.

R/rvar-cast.R
mjskay: Normally I'd be okay with this change, but for these vctrs definitions I've found it helpful to have them laid out like this when editing them --- it's easier to keep them consistent across the various combinations of definitions that should follow similar patterns. If it's alright without you I'd like to leave them this way in this file (I am adjusting line-spacing everywhere else).

R/rvar-dist.R
```diff
@@ -59,14 +71,42 @@ cdf.rvar <- function(x, q, ...) {
   })
 }
 
+#' @rdname rvar-dist
+#' @export
+cdf.rvar_factor <- function(x, q, ...) {
+  stop_no_call("cdf() cannot be calculated for rvar_factor() objects.")
```
mjskay, line 77: Note to self: to be consistent with distributional::cdf() and ::quantile() on categorical distributions, these should probably return `NA` vectors instead of an error. My experience adding rvar_factor support to ggdist also suggests this would be easier to use than throwing an error as well.

R/as_draws_array.R
```diff
@@ -92,6 +97,11 @@ as_draws_array.draws_rvars <- function(x, ...) {
     return(empty_draws_array(variables(x)))
   }
 
+  factor_variables <- vapply(x, is_rvar_factor, logical(1))
```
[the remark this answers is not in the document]
mjskay: the conversion happens in the next step: `draws <- do.call(`cbind`, ...`. `cbind` drops class information from vectors, so factors are converted to numeric. It is a bit obscure (but documented) behavior, so I'll add a comment noting it.

Re: having a separate function for the check+warning, good idea.

R/as_draws_rvars.R
mjskay: like, a generic function that does this for any draws type?

R/rvar-dist.R
mjskay: done

R/rvar-factor.R
mjskay: good catch, thanks! done

tests/testthat/test-rvar-apply.R
mjskay: Good idea. Adding a simple test (not a whole new test case) to check if the ordered property is preserved (if it works in one sub-case it should in the other)

mjskay:
> This PR looks already really good, up to a few minor things, mostly style related.

Thanks for the really thorough review! I've addressed everything except two comments (see those replies).

> Some of the rvar internals remain a bit of black magic to me, so I don't understand every detail of it, but I trust you and the tests you put in place on that one.

Heh, yeah, I agree it is pretty arcane. Getting rvar to mimic everything it does so far has been quite a journey (base R datatypes are... certainly something!), and I should probably clean up and document the rvar internals a bit more with everything I've had to figure out to get it to work (plus I sometimes forget when I've been away from it a bit).

R/as_draws_rvars.R
mjskay: (I think I'm not seeing where to use it currently, so I'm not sure exactly the signature / definition)

paul-buerkner:
Thank you for your changes and explanations of what you what to keep the way it is. It does all make sense to me. I want to give people a bit more time to check out the new features and perhaps add a review. So I will put a reminder in my calender for start of next week to merge this PR if no new things have come up until then. Would that sound good to you?

mjskay:
Sounds great, thanks!

Should we ping some folks who might be able to provide feedback? Off the top of my head: @isabellaghement, @bwiernik, @jgabry, @avehtari? If any of you have a chance to "kick the tires" of `rvar_factor()` / `rvar_ordered()` on this branch sometime in the next week, it would be much appreciated (of course, no pressure!).

avehtari:
pinging @TeemuSailynoja as he has been running recently some categorical and ordinal models

bwiernik:
Try to run some tests with it this weekend

paul-buerkner:
@mjskay From my side, this is ready to merge. Anything you want to change before I merge?

mjskay:
Sorry for the delayed response - yes, I think this is good for now. I've built some stuff in ggdist on top of it and haven't run into anything new that needs changing, so that's the best I can say at the moment without other folks testing it. Thanks!

paul-buerkner:
Lovely, merging now!

paul-buerkner (approved):

mjskay:
Thanks!

PR #272, opened 2023-02-03
Do not automatically create an rvar_factor when rvar() is passed a numeric array with levels

mjskay:
#### Summary

This is a fix for a minor issue with `rvar_factor()` I've run into after using a bit more in practice. 

Currently, given an array `x` with a `"levels"` attribute, `rvar(x)` will attempt to construct an `rvar_factor` instead of a numeric `rvar`. However, this can fail if `x` has more unique values than are in `"levels"` --- basically, if `"levels"` is not intended as a set of levels corresponding to integer entries of `x`. An example of such objects includes some output from `brms::posterior_predict()`, which makes `rvar(brms::posterior_predict())` fail in some cases (when it should really always work).

This PR is just a slight change: instead of automatically attempting to use the `"levels"` attribute in `rvar(x)`, this makes it so that the `"levels"` attribute is only examined if the user invokes `rvar_factor(x)`. This way, the user effectively has to declare that the input should be treated as factor-like instead of us trying to infer that.

Other types that are unambiguously factor-like (specifically, character arrays, factors, and ordereds) will still produce `rvar_factor`s if passed to `rvar()`.

#### Copyright and Licensing

### DIFF
--- a/R/rvar-.R
+++ b/R/rvar-.R
@@ -806,22 +806,11 @@ cleanup_rvar_draws <- function(x) {
     rownames(x) <- as.character(seq_rows(x))
   }
 
-  # if x is factor-like (character or with "levels" attr), make it a factor
-  x <- as_factor_if_factor_like(x)
-
-  x
-}
-
-#' convert factor-like objects (character vectors or numerics with "levels"
-#' attributes) to factors. Leaves non-factor-like objects alone.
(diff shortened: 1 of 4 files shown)


paul-buerkner:
Thanks, looks good! Merging now.

PR #293, opened 2023-06-14
Slicing rvars with rvars (and rvar_ifelse)

### LINKED ISSUE #282: Subsetting an rvar with an rvar
@mjskay I absolutely love the `rvar` class! So much better than dealing with pesky vectors / matrices / arrays!

Is there a way to subset an `rvar` with an `rvar`?

If not, it would be great to be able to do something like this:

```R
library(posterior)

A <- rvar(rnorm(100))
B <- rvar(rnorm(100))

(idx <- A > B)
#> rvar<100>[1] mean ± sd:
#> [1] 0.44 ± 0.5

Pr(idx) # is logical
#> [1] 0.44

A[idx] # make a new rvar based only on the subset of samples for which idx is true
#> Error...
```

This would really be great for order restrictions (e.g., the type used in [C14 carbon dating](https://arxiv.org/pdf/1704.07141.pdf)...)

mjskay:
#### Summary

Finally submitting a PR to close #282. It:

- adds `ifelse.rvar(test, yes, no)`, which broadcasts `test`, `yes`, and `no` (all rvars or castable to rvars) to a common shape and then does a vectorized ifelse on them.
- `x[i]` where `i` is a scalar logical rvar, which would return an rvar of the same shape as `x` but with draws subset by the logical draws of `i`. It also implements the assignment form `x[i] <- value`.
- `x[[i]]` where `i` is a scalar numeric rvar, which would return a scalar rvar. If $x_{(n)}$ is the $n$-th draw of `x`, $y_{(n)}$ would be $x_{(n)}[[i_{(n)}]]$. It also implements the assignment form `x[[i]] <- value`.

While I was in here, I also fixed a minor bug in correctness for some vctrs functions applied to rvars (e.g. `vctrs::vec_identify_runs(<rvar>)`) by implementing `vec_proxy_equal(<rvar>)`.

@mattansb your thoughts are appreciated since your suggestion precipitated the above changes.

#### Copyright and Licensing


paul-buerkner:
Thank you @mjskay. Waiting for review and mergine to the response of @mattansb.

mattansb:
Hey @mjskay this looks great!

1. I couldn't find `ifelse.rvar()` only `rvar_ifelse()` - is that what you meant?

```R
# installed today from 
# remotes::install_github("stan-dev/posterior@rvar_slice")
library(posterior) 

yes <- rvar(c(1, 1, 1, 1))
no <- rvar(c(2, 2, 2, 2))
test <- rvar(c(T, F, T, T))

ifelse(test, yes, no)
#> Error in `as.logical()`:
#> ! Can't convert `x` <rvar<4>> to <logical>.
#> Run `rlang::last_trace()` to see where the error occurred.

rvar_ifelse(test, yes, no)
#> rvar<4>[1] mean ± sd:
#> [1] 1.2 ± 0.5 

draws_of(.Last.value)
#>   [,1]
#> 1    1
#> 2    2
#> 3    1
#> 4    1
```

2. `[.rvar` is working great!

3. I can't figure out what `[[.rvar` is doing.  Am I using this wrong?

```R
x <- rvar(5:8)
i <- rvar(c(1,1,4,4))

x[[i]]
#> rvar<4>[1] mean ± sd:
#> [1] NA ± NA
```

mjskay:
> 1. I couldn't find `ifelse.rvar()` only `rvar_ifelse()` - is that what you meant?

Yeah sorry, that's what I meant.



> 2\. `[.rvar` is working great!

Great!



> 4\. I can't figure out what `[[.rvar` is doing.  Am I using this wrong?
> 
> ```r
> x <- rvar(5:8)
> i <- rvar(c(1,1,4,4))
> 
> x[[i]]
> #> rvar<4>[1] mean ± sd:
> #> [1] NA ± NA
> ```

Ah yeah, that's the expected output for what you gave it, but that's not the intended use. `x[[<rvar>]]` is intended as a generalization of `x[[i]]` where `i` is a scalar numeric --- so e.g. `x[[rvar(1)]]` should be equivalent to `x[[1]]` since `rvar(1)` is a constant. So just like `x[[<numeric>]]` always returns a scalar (length-1) rvar, so does `x[[<rvar>]]`, it just uses a different value for the index in each draw.

So in the first two draws in your example you are grabbing the value of `x[[1]]`, but in the third and fourth draw you are grabbing `x[[4]]`, which is `NA` because `x` has length 1. So your result looks like this:

```r
draws_of(x[[i]])
#>   [,1]
#> 1    5
#> 2    6
#> 3   NA
#> 4   NA
```

Here's an example of using all three methods to make a mixture distribution that might clarify things.

First, logical indexing:

``` r
component = rvar_rng(rnorm, 2, mean = c(1, 5))
component
#> rvar<4000>[2] mean ± sd:
#> [1] 0.99 ± 1  4.99 ± 1

i = rvar_rng(rbinom, 1, size = 1, p = 0.75) + 1
i
#> rvar<4000>[1] mean ± sd:
#> [1] 1.7 ± 0.44

x = component[[1]]
x[i == 2] = component[[2]][i == 2]
x
#> rvar<4000>[1] mean ± sd:
#> [1] 4 ± 2

ggplot() +
  ggdist::stat_slab(aes(xdist = x))
```



You could create the same thing using `rvar_ifelse()`:

```r
x = rvar_ifelse(i == 1, component[[1]], component[[2]])
x
#> rvar<4000>[1] mean ± sd:
#> [1] 4 ± 2
```

Or, since `i` is a scalar numeric rvar whose values are either `1` or `2` within each draw, you can use it as an index directly on `component` to create the mixture:

```r
x = component[[i]]
x
#> rvar<4000>[1] mean ± sd:
#> [1] 4 ± 2
```

This approach is also nice because it generalizes easily to more than two components.

Anyway, this makes me think it would be worth creating a documentation page for `[.rvar` and `[[.rvar`, and maybe put some examples like above into the `rvar` vignette. I will do that on this PR before we merge it.

paul-buerkner:
I think some additional doc would help. I really like these semantics but they require some time to understand and use intuitively. Thank you so much for your work on this!

mattansb:
Thanks @mjskay - I think I got the mental model now. 
I also now understand why `[[.ravr` will always return a scalar `rvar` - pretty cool!

Would be happy to review any additional docs.

mjskay:
Great, thanks both! I will probably make a doc page that describes all the different types of slicing for rvars, including slicing with and without rvars. I'll also adds some examples to the vignette. I'll let you all know here when that's done.

mjskay:
I've added a help page on slicing (`help("rvar-slice")`) as well as some examples of slicing to `vignette("rvar")`, along with a simple example of visualizing the result (with the help of ggdist --- so I had to add it to Suggests as well). 

Feedback welcome. Once we're happy with the docs I think this is ready to merge.

paul-buerkner:
Thanks! I don't have time to check it right now so I would appreciate if others would check.

mattansb:
Looks great @mjskay !

I've already started using rvar slicing, and I mostly use it to subsetting posteriors according to some condition *of the posterior*. For example:

``` r
library(posterior)
#> This is posterior version 1.4.1.9000
#> 
#> Attaching package: 'posterior'
#> The following objects are masked from 'package:stats':
#> 
#>     mad, sd, var
#> The following objects are masked from 'package:base':
#> 
#>     %in%, match
library(ggplot2)
#> Warning: package 'ggplot2' was built under R version 4.2.3
library(ggdist)

component = rvar_rng(rnorm, 2, mean = c(1, 2), sd = c(1, 2))

i <- component[2] > component[1] # Find where the 2nd dist is larger than the 1st

component_s <- component[i]

ndraws(component)
#> [1] 4000
ndraws(component_s)
#> [1] 2659

ggplot() +
  stat_slab(aes(fill = "component_s", xdist = component_s), alpha = 0.8) +
  stat_slab(aes(fill = "component", xdist = component), alpha = 0.8) +
  scale_thickness_shared()
```



<sup>Created on 2023-06-19 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>


Maybe something like this can be included in the vignette?

paul-buerkner:
@mjskay do you want to do any changes before I merge?

mjskay:
Sorry, got distracted by other stuff :)

@paul-buerkner I think this is good for now, I'd say merge it?

@mattansb I don't know if I want to overwhelm the vignette with several examples of one operator since there's already a lot going on in there, so for now I'd rather leave it as is. Thanks for the feedback!

mattansb:
Then you leave me no choice - but to write about it in a blog post! 😉

Thank _you_!

mjskay:
> Then you leave me no choice - but to write about it in a blog post! 😉

Perfect!! 😊

paul-buerkner:
Great, merging now!

PR #296, opened 2023-08-01
add set_variables alias of variables<-

### LINKED ISSUE #281: [feature] set_variable_names (alias of variables<-) for piping
In many cases, one has unnamed draws, and it could be useful to have a `set_variable_names` function (equivalent to `magrittr::set_col_names`) which could be used in piping workflows. This could be particularly useful when using prediction functions that return unnamed draws.

Functionally this would work like `variables<-` but would allow for piping.

For example
```{r}
unnamed_draws|> 
    set_variable_names(c("m1", "m2")) # here we set all variable names at once
```
equivalent to
```
variables(unnamed_draws) <- c("m1", "m2")
```

n-kall:
#### Summary

Adds `set_variables` function which is an alias of `variables<-`. Fixes #281

Example functionality:

```{r}
x <- as_draws(matrix(rnorm(100), ncol = 2))
variables(x)
#> [1] "...1" "...2"

x |> set_variables(c("theta[1]", "theta[2]")) |> variables()
#> [1] "theta[1]" "theta[2]"

# this is equivalent to
variables(x) <- c("theta[1]", "theta[2]")
variables(x)
#> [1] "theta[1]" "theta[2]"

```

#### Copyright and Licensing


paul-buerkner:
Thank you! Would you mind moving the code to the R file where variables() <- is defined to keep them together? Same with the tests.

n-kall:
``

> Would you mind moving the code to the R file where variables() <- is defined to keep them together? Same with the tests.

Done!

R/draws-index.R
```diff
@@ -141,6 +141,42 @@ remove_reserved_variable_names <- function(variables, reserved) {
...
+#'
+#' @return Returns a [`draws`] object of the same format as `.x`, with
+#'   variables named as specified.
+#'
+#' @seealso [`variables`]
+#' @examples
+#' x <- as_draws(matrix(rnorm(100), ncol = 2))
+#' variables(x)
+#'
+#' x |> set_variables(c("theta[1]", "theta[2]")) |> variables()
```
paul-buerkner, line 161: Please avoid the use of the pipe in the docs because this may lead to a failur in old R versions not having the pipe.

R/draws-index.R
```diff
@@ -141,6 +141,42 @@ remove_reserved_variable_names <- function(variables, reserved) {
...
+#'
+#' # this is equivalent to
+#' variables(x) <- c("theta[1]", "theta[2]")
+#' variables(x)
+#'
+#' @export
+set_variables <- function(.x, ...) {
```
paul-buerkner, line 168: why use .x as the first argument here? Wouldn't x be okay too?

paul-buerkner, line 168: shall the generic enforce the second argument be called "variables"? right now the generic doesn't specify that so other methods could in theory call the argument differently which might end up being confusing, I think.

paul-buerkner (changes requested):
Thank you! I have a few minor comments before merging this.

R/draws-index.R
n-kall: Fixed by changing variable to `x`

n-kall: This was an oversight, fixed now so that the generic has a variables argument

R/draws-index.R
n-kall: Fixed

R/draws-index.R
```diff
@@ -141,6 +141,43 @@ remove_reserved_variable_names <- function(variables, reserved) {
...
+#' variables(x)
+#'
+#' @export
+set_variables <- function(x, variables, ...) {
+  UseMethod("set_variables")
+}
+
+#' @rdname set_variables
+#' @export
+set_variables.draws <- function(x, variables, ...) {
```
mjskay, line 175: Why make `set_variables` generic? If it is a non-generic function that just forwards to `variables<-` then implementers of other classes need only override the generic there.

n-kall: I didn't think too much about this either way, as I just used `rename_variables` as a template, which is a generic with only a `rename_variables.draws` method defined. If it's better to change it to a non-generic function, I can do that

mjskay: I defer to @paul-buerkner, but IMO it would be simpler :).

jgabry: non-generic sounds good to me too

n-kall: I changed it to non-generic now

R/draws-index.R
```diff
@@ -141,6 +141,37 @@ remove_reserved_variable_names <- function(variables, reserved) {
...
+#'
+#' @return Returns a [`draws`] object of the same format as `x`, with
+#'   variables named as specified.
+#'
+#' @seealso [`variables`]
+#' @examples
+#' x <- as_draws(matrix(rnorm(100), ncol = 2))
+#' variables(x)
+#'
+#' set_variables(x, c("theta[1]", "theta[2]"))
```
paul-buerkner, line 161: you are not assigning to x here. Is that intentional?

paul-buerkner (changes requested):
looks good. One minor thing before I will merge

R/draws-index.R
n-kall: Good catch! Fixed this now.

paul-buerkner (approved):

paul-buerkner:
Thank you for adding this!

n-kall:
> Thank you for adding this!

Thanks for the review!

PR #301, opened 2023-10-01
Add examples of converting base R 3-D array and mcmc/mcmc.list objects to README.md

avehtari:
Add examples of converting base R 3-D array and mcmc/mcmc.list objects

As it took me a moment to figure out how to convert R 3-D array to draws so that variables are named (as the approach used for matrix doesn't work) it is probably helpful for others to see the example, too.

I didn't even know that mcmc/mcmc.list object conversion is supported, and it would be useful to mention it at least once somewhere, so that it is possible to point others to tha information.

### DIFF
--- a/README.Rmd
+++ b/README.Rmd
@@ -222,6 +222,27 @@ Instead of `as_draws_matrix()` we also could have just used `as_draws()`, which
 attempts to find the closest available format to the input object. In this case
 this would result in a `draws_matrix` object either way.
 
+The above matrix example contained only one chain. Multi-chain draws could be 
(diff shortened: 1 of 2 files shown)


paul-buerkner:
We do not require coda and rjags in posterior so I assume we will fail some checks by adding these packages to the readme. Did you run the usual CMD checks with this already? if not, I will do it when working on the new posterior release in the next few days.

paul-buerkner:
I think you edited the readme.**md** file directly, but all changes should only go to the readme.**Rmd** file which will then be used to render the corresponding .md file. Anyway, I have made the necessary changes and will merge now. Thank you for this addition!

PR #318, opened 2023-11-17
New helper function repair_variable_names

### LINKED ISSUE #317: Add a `repair_variable_names` function
This allows you to translate from the names Stan uses (like `theta.1`) to those Posterior wants (`theta[1]`).

This function exists (privately) in cmdstanr, and is currently very simple:

```R
repair_variable_names <- function(names) {
    names <- sub("\\.", "[", names)
    names <- gsub("\\.", ",", names)
    names[grep("\\[", names)] <- paste0(names[grep("\\[", names)], "]")
    names
}
```

This doesn't work for complex numbers or tuples, but since posterior doesn't seem to support those either I think this is fine

jgabry:
closes #317

#### Summary

`repair_variable_names` converts names using periods (`theta.1.1`) to square brackets (`theta[1,1])

#### Copyright and Licensing


R/draws-index.R
```diff
@@ -173,6 +175,31 @@ set_variables <- function(x, variables, ...) {
   return(x)
 }
 
+#' @rdname set_variables
+#' @param old_variables (character) Variable names to repair. Should be variable
```
jgabry, line 179: For now I went with `old_variables` as the argument name, but happy to change this based on feedback.

R/draws-index.R
```diff
@@ -173,6 +175,31 @@ set_variables <- function(x, variables, ...) {
...
+#' repair_variable_names(colnames(x))
+#' x <- set_variables(as_draws(x), repair_variable_names(colnames(x)))
+#' variables(x)
+#'
+#' @export
+repair_variable_names <- function(old_variables) {
+  if (!all(grepl("\\.", old_variables))) {
+    stop_no_call(
+      "All names in 'old_variables' must contain at least one '.' in the name."
+    )
```
jgabry, line 195: For simplicity I added an error if all the names don't have a `.` in it. It would be strange to pass in a vector like `c("theta.1", "theta[2]")` so I don't think we need to write out the logic to handle mixed cases like that. But let me know if you disagree and I can add that logic.

jgabry:
Maybe `repair` isn't the best name actually? @avehtari suggested `convert` in https://github.com/stan-dev/posterior/issues/317#issuecomment-1816897197. I'm open to other suggestions too.

paul-buerkner:
I like "repair". Having the dots instead of [] is kind of a thing we don't really want. So we are repairing them to look reasonable again.

R/draws-index.R
WardBrian: This logic seems very unhelpful if I have some variables which are scalars. I can no longer just pass the entire list to this function 

I believe the set of transformations here is idempotent, so if someone passes `theta[1]`, nothing will be changed, so why not allow it

mjskay:
Re: function name: is the intention to create a function that specifically converts from Stan names, or is the intention that this function might (in the future) support other non-posterior-friendly naming conventions? I think that should influence the naming; e.g. is this "convert_from_stan_names" (though perhaps less verbose) or the more generic "repair_variable_names" or "convert_variable_names"? Or is it "repair_dotted_indices" or "repair_dotted_index_names" or something like that?

jgabry:
Good point @mjskay. If this will only ever be used to convert Stan names we could include that in the function name. What do others think?

avehtari:
I think it would be best to make the function specific for Stan naming conventions, and then add other functions if we add support for other packages. The name "convert_from_stan_names" would be then very descriptive, and I don't know how to make it shorter

Furthermore, in #317 @WardBrian wrote
> Complex variables use .real and .imag as their CSV names in Stan. So something like a complex_vector[2] foo has names foo.1.real,foo.1.imag,foo.2.real,foo.2.imag
>
> Tuples use : to separate "slots" in a tuple. So a tuple(real, int) a is printed as a:1,a:2. This stacks naturally with the . for arrays of tuples/tuples of arrays.

If we want to support these, too, I assume that we would like to have something like
"foo.1.real" -> "foo_real[1]"
and
"a:1" -> "a_1"

And the natural complex type support is a separate issue #319 and PR #321

paul-buerkner:
I agree with Jonah and Aki.

WardBrian:
If you did want to support tuple names, this essentially just needs to split on `:` and loop over the resulting parts. See the code that does something similar in Stan: https://github.com/stan-dev/stan/blob/667d5b2d7c5ea1072011919a1acd76a9d3061d01/src/stan/io/stan_csv_reader.hpp#L16

PR #321, opened 2023-11-20
Complex numbers

### LINKED ISSUE #319: Support for complex numbers
After seeing @WardBrian's comment here: https://github.com/stan-dev/posterior/issues/317#issuecomment-1817035410

it occurred to me to check our support for complex numbers. It appears none of the existing formats support them. 

Support for converting complex numbers that are encoded as arrays with a dimension that has a "real" and "complex" component *would* be straightforward if our underlying types worked with R's complex number data type; e.g. given a variable like this:

```r
> cbind(real = rvar_rng(rnorm, 5), imag = rvar_rng(rnorm, 5, 1))
rvar<4000>[5,2] mean ± sd:
     real            imag           
[1,] -0.0069 ± 1.01   0.9964 ± 1.02 
[2,] -0.0093 ± 0.99   0.9854 ± 0.98 
[3,] -0.0061 ± 1.02   0.9984 ± 0.99 
[4,]  0.0301 ± 0.99   0.9650 ± 1.00 
[5,]  0.0030 ± 1.00   0.9919 ± 0.99 
```

The following would be an easy way to turn the above into a complex rvar backed by the base complex number data type if it wasn't generating a bunch of warnings:

```r
> x = cbind(real = rvar_rng(rnorm, 5), imag = rvar_rng(rnorm, 5, 1))
> x[,"real"] + x[,"imag"] * rvar(1i)
rvar<4000>[5,1] mean ± sd:
     real            
[1,] -0.02+0.99i ± 1 
[2,]  0.00+1.00i ± 1 
[3,]  0.00+1.02i ± 1 
[4,]  0.00+1.00i ± 1 
[5,]  0.02+1.00i ± 1 
Warning messages:
1: In var(if (is.vector(x) || is.factor(x)) x else as.double(x), na.rm = na.rm) :
  imaginary parts discarded in coercion
2: In var(if (is.vector(x) || is.factor(x)) x else as.double(x), na.rm = na.rm) :
  imaginary parts discarded in coercion
3: In var(if (is.vector(x) || is.factor(x)) x else as.double(x), na.rm = na.rm) :
  imaginary parts discarded in coercion
4: In var(if (is.vector(x) || is.factor(x)) x else as.double(x), na.rm = na.rm) :
  imaginary parts discarded in coercion
5: In var(if (is.vector(x) || is.factor(x)) x else as.double(x), na.rm = na.rm) :
  imaginary parts discarded in coercion
```

The other draws types all exhibit similar issues. Should be solvable though I think.

mjskay:
#### Summary

This is a draft PR to support complex numbers in draws and rvar data types. It closes #319. The main changes, plus some points of conversation:

- Adjust two checks in as_draws.R to allow complex values (this is the main thing that buys us support in draws formats)
- Provide an implementation of `sd()`, `variance()`, and `var()` for complex values. This fixes printing issues, especially with `rvar`s. The implementation of `variance()` follows the definition of the [variance of a complex random variable](https://en.wikipedia.org/wiki/Complex_random_variable#Variance_and_pseudo-variance) as the sum of the variances of its real and imaginary components; and its sd is taken as the square root of this.
- Ensure other rvar functions work with complex numbers, or (in cases of functions that should not work with complex numbers), raise an error. Functions that do not work with complex numbers include:
   - [rvar_]min/max/range - since complex numbers are not orderable, these are not defined
   - [rvar_]median - base R will compute a median of a complex number by taking the medians of each component; however, I'm not sure sure if that's correct (or if something like a geometric median would make more sense). One might argue that if someone just wanted the point estimates of the medians for each component, as in `summarise_draws()`, the base R approach could make sense. Conservatively I've let this throw an error, but would be curious thoughts.
   - Base R `quantile()` on complex numbers seems (like median) to just treat each component separately, which I think is wrong, but again might be okay for us (in `rvar_quantile` and `[rvar_]quantile2`) if people just want the separate point estimates for the two components. I'm not sure.
   - [rvar_]density, [rvar_]cdf - if implemented, these should (I think) be joint densities/cdfs for the two components, i.e. using 2d ecdfs / 2d density estimators. Absent that implementation, for now, these throw an error.
- I'm not sure what to do about convergence metrics. At the moment, most fail because of a check in `is_constant()` that requires `min` / `max`, which are not defined on complex numbers. However, some succeed but I think give incorrect answers: e.g. `rhat()` appears to work, but I think returns incorrect values, as it relies on `rank()` in `z_scale()`, but base R's `rank()` ranks complex numbers in lexicographic order, when in reality it should probably return an error. I'm not sure if there's a good definition of any of these convergence metrics for complex numbers (maybe something like max rhat of both components?) or if these should all just return errors.

The main motivating example for this, inspired by @WardBrian's comment, does work now:

```r
> x = cbind(real = rvar_rng(rnorm, 5), imag = rvar_rng(rnorm, 5, 1))
> x[,"real"] + x[,"imag"] * 1i
rvar<4000>[5,1] mean ± sd:
     real              
[1,]  0.00+0.98i ± 1.4 
[2,] -0.01+0.98i ± 1.4 
[3,]  0.01+0.97i ± 1.4 
[4,]  0.00+1.01i ± 1.4 
[5,]  0.01+0.99i ± 1.4
```

#### Copyright and Licensing


WardBrian:
Does this handle the automatic re-shaping if I pass something that has `colnames` like `x.1.real,...`? Note that it is slightly subtle to do this correctly, since you end up with essentially strided draws in most ways of arranging these

mjskay:
If I understand correctly, you're asking if it matters where in the order of dimensions the dimension indexing the real and imaginary parts is? I don't think it should make a difference; array slicing should handle that.

e.g. here's an rvar where the second dimension indexes the real/imaginary parts:

``` r
x = rvar(array(
  1:96, dim = c(4,4,2,3), 
  dimnames = list(NULL, NULL, c("real", "imag"), NULL)
))
x
#> rvar<4>[4,2,3] mean ± sd:
#> , , 1
#> 
#>      real        imag       
#> [1,]  2.5 ± 1.3  18.5 ± 1.3 
#> [2,]  6.5 ± 1.3  22.5 ± 1.3 
#> [3,] 10.5 ± 1.3  26.5 ± 1.3 
#> [4,] 14.5 ± 1.3  30.5 ± 1.3 
#> 
#> , , 2
#> 
#>      real        imag       
#> [1,] 34.5 ± 1.3  50.5 ± 1.3 
#> [2,] 38.5 ± 1.3  54.5 ± 1.3 
#> [3,] 42.5 ± 1.3  58.5 ± 1.3 
#> [4,] 46.5 ± 1.3  62.5 ± 1.3 
#> 
#> , , 3
#> 
#>      real        imag       
#> [1,] 66.5 ± 1.3  82.5 ± 1.3 
#> [2,] 70.5 ± 1.3  86.5 ± 1.3 
#> [3,] 74.5 ± 1.3  90.5 ± 1.3 
#> [4,] 78.5 ± 1.3  94.5 ± 1.3
```

Showing it as a `draws_df` shows the corresponding variable names in most formats:

```r
as_draws_df(x)
#> # A draws_df: 4 iterations, 1 chains, and 24 variables
#>   x[1,real,1] x[2,real,1] x[3,real,1] x[4,real,1] x[1,imag,1] x[2,imag,1]
#> 1           1           5           9          13          17          21
#> 2           2           6          10          14          18          22
#> 3           3           7          11          15          19          23
#> 4           4           8          12          16          20          24
#>   x[3,imag,1] x[4,imag,1]
#> 1          25          29
#> 2          26          30
#> 3          27          31
#> 4          28          32
#> # ... with 16 more variables
#> # ... hidden reserved variables {'.chain', '.iteration', '.draw'}
```

We can still just slice the two components and sum them. We could also do `x[,"real",] + x[,"imag",] * 1i` (note the extra commas), but it's not necessary as `rvar` fills in missing dimensions on the end during a slice:

```r
x[,"real"] + x[,"imag"] * 1i
#> rvar<4>[4,1,3] mean ± sd:
#> , , 1
#> 
#>      real         
#> [1,]  2+18i ± 1.8 
#> [2,]  6+22i ± 1.8 
#> [3,] 10+26i ± 1.8 
#> [4,] 14+30i ± 1.8 
#> 
#> , , 2
#> 
#>      real         
#> [1,] 34+50i ± 1.8 
#> [2,] 38+54i ± 1.8 
#> [3,] 42+58i ± 1.8 
#> [4,] 46+62i ± 1.8 
#> 
#> , , 3
#> 
#>      real         
#> [1,] 66+82i ± 1.8 
#> [2,] 70+86i ± 1.8 
#> [3,] 74+90i ± 1.8 
#> [4,] 78+94i ± 1.8
```

If you don't want the (now-useless) "real" dim, you can drop it:

```r
drop(x[,"real"] + x[,"imag"] * 1i)
#> rvar<4>[4,3] mean ± sd:
#>      [,1]          [,2]          [,3]         
#> [1,]  2+18i ± 1.8  34+50i ± 1.8  66+82i ± 1.8 
#> [2,]  6+22i ± 1.8  38+54i ± 1.8  70+86i ± 1.8 
#> [3,] 10+26i ± 1.8  42+58i ± 1.8  74+90i ± 1.8 
#> [4,] 14+30i ± 1.8  46+62i ± 1.8  78+94i ± 1.8
```

Which also looks right if we again convert to a format like draws_df:

```r
as_draws_df(drop(x[,"real"] + x[,"imag"] * 1i))
#> # A draws_df: 4 iterations, 1 chains, and 12 variables
#>   x[1,1] x[2,1] x[3,1] x[4,1] x[1,2] x[2,2] x[3,2] x[4,2]
#> 1  1+17i  5+21i  9+25i 13+29i 33+49i 37+53i 41+57i 45+61i
#> 2  2+18i  6+22i 10+26i 14+30i 34+50i 38+54i 42+58i 46+62i
#> 3  3+19i  7+23i 11+27i 15+31i 35+51i 39+55i 43+59i 47+63i
#> 4  4+20i  8+24i 12+28i 16+32i 36+52i 40+56i 44+60i 48+64i
#> # ... with 4 more variables
#> # ... hidden reserved variables {'.chain', '.iteration', '.draw'}
```

Issue #127, opened 2021-05-06
Feature: Automate thinning


TeemuSailynoja:
Per discussion with @avehtari , it would be good to provide functionality to automate thinning of samples with autocorrelation using the `thin` method and among other strategies, the lags mentioned in Issue #33.

Possible thinning strategies: 
- Thin by the Geyer's truncation lag "Conservative thinning"
- 95% or user defined percentage of the cumulative autocorrelation time "Flexible thinning"
- Thin by ceiling( N/N<sub>eff</sub> ) as suggested in [5.1. of the SBC paper](https://arxiv.org/pdf/1804.06788.pdf)
- Thin according to the quantile (5%, 10%, ..., 95%) giving the largest thinning lag with the above method, also suggested in the SBC paper. "quantile based thinning"
- Thin to the lag  2m  s. t. p<sub>2m</sub> + p<sub>2m+1</sub> < 0, where p<sub>2m</sub> is the autocorrelation with lag 2m. "Aggressive thinning"

Functionality:
- By default, recommend a thinning strategy, return thinned sample and provide feedback of the results (remaining samples per chain, remaining autocorrelations, resulting thinning lag).
- With user specified thinning strategy, return the thinned sample and provide feedback.

Thoughts:
- Can the thinning be done iteratively, which criteria to set for the thinning to be adequate?
- Drawbacks of starting from the conservative thinning and using a less strict strategy, if the number of remaining samples would otherwise be too low?

avehtari:
Closed by #326

PR #326, opened 2023-12-11
add automatic thinning of draws

n-kall:
#### Summary

After discussions with @avehtari this PR adds automatic thinning of draws based on ESS as suggested by [Säilynoja et al. 2022](https://link.springer.com/article/10.1007/s11222-022-10090-6) (appendix). 

Thin by = S / (min(ess_tail, ess_bulk)). For multiple chains, this is ndraws_per_chain / mean_over_chains(min(ess_tail, ess_bulk))

Addresses issue #127 
#### Copyright and Licensing


n-kall:
@avehtari Should it be ceiling() or round(). In the issue #127 it says ceiling, but I don't remember what we decided

avehtari:
How about thinning with non-integers? E.g. `round(seq(1,100,by=max(1,1.5)))` gives valid indeces

R/thin_draws.R
```diff
@@ -4,8 +4,11 @@
 #'
 #' @aliases thin
 #' @template args-methods-x
-#' @param thin (positive integer) The period for selecting draws.
+#' @param thin (positive integer) The period for selecting draws. If
+#'   omitted, this will be automatically calculated based on bulk-ESS
```
mjskay, line 8: Minor nitpick, this should say `NULL` not "omitted" (which could be taken to mean a missing arg):

```suggestion
#'   `NULL`, this will be automatically calculated based on bulk-ESS
```

n-kall:
I've changed the thinning to handle non-integer values. I think this is now ready for further review

paul-buerkner:
Looks good to me. @avehtari  do you have any comments? Otherwise, I will merge.

avehtari:
It would be good that the documentation would explicitly mention what happens with non-integer thin values. Even if people would not use non-integer as argument, if they use automatic thinning, they may be surprised by the number of returned iterations unless behavior of non-integer thin is documented

n-kall:
@avehtari how about this for the doc:
"thin (numeric) The period for selecting draws. Must be between 1 and the number of iterations. If the value is not an integer, the draws will be selected such that average interval between them approaches the thin value. If `NULL`, it will be automatically calculated based on bulk and tail effective sample size as suggested by Säilynoja et al. (2022)."

n-kall:
Updated the documentation based on dicussions with @avehtari. I think it's ready now

avehtari:
OK for me

Issue #184, opened 2021-07-23
Improve support (or documentation) for weighted draws


avehtari:
`?weight_draws` says `Add weights to draws objects, with one weight per draw, for use in subsequent weighting operations.` Currently the only supported subsequent operation is `resample_draws()`, but the documentation doesn't make it clear. Ideally `rvars` support would be extended to support weighted draws, but if that is infeasible, the documentation should be clarified.

Example:
```
> x = rvar_rng(rnorm, 4, mean = 1:4, sd = 2)
> w=runif(4000)
```
`weight_draws()` is not defined for draws_rvars:
```
> xw = weight_draws(x, w, log=FALSE)
Error in UseMethod("weight_draws") : 
  no applicable method for 'weight_draws' applied to an object of class "c('rvar', 'vctrs_vctr', 'list')"
```
There is no error if weighted draws_df is converted to draws_rvars:
```
> xw = as_draws_rvars(weight_draws(as_draws_df(x), w, log=FALSE))
```
The default print method fives the unweighted result
```
> xw
# A draws_rvars: 4000 iterations, 1 chains, and 1 variables
$x: rvar<4000>[4] mean ± sd:
[1] 0.96 ± 2  1.95 ± 2  3.02 ± 2  4.05 ± 2 
```
and mean fails
```
> mean(xw)
[1] NA
Warning message:
In mean.default(xw) : argument is not numeric or logical: returning NA
```

mjskay:
Thanks, I think there's a couple of different issues here.

> weight_draws() is not defined for draws_rvars

It is, I think you were using it on an `rvar` not a `draws_rvars` object. Based on your example:

```r
> x = rvar_rng(rnorm, 4, mean = 1:4, sd = 2)
> w = runif(4000)
> x
rvar<4000>[4] mean ± sd:
[1] 1 ± 2  2 ± 2  3 ± 2  4 ± 2 
> 
> d = draws_rvars(x = x)
> d
# A draws_rvars: 4000 iterations, 1 chains, and 1 variables
$x: rvar<4000>[4] mean ± sd:
[1] 1 ± 2  2 ± 2  3 ± 2  4 ± 2 

> 
> weight_draws(d, w)
# A draws_rvars: 4000 iterations, 1 chains, and 1 variables
$x: rvar<4000>[4] mean ± sd:
[1] 1 ± 2  2 ± 2  3 ± 2  4 ± 2 

# ... hidden reserved variables {'.log_weight'}
```

[...]

PR #331, opened 2024-01-06
Weighted rvars

mjskay:
### Summary

This PR aims to address (at least part of) #184 by implementing weighted `rvar`s.

Currently, `rvar`s cannot contain weights, and weighting of them can only be done by putting them in a `draws_rvars` object that itself contains a `".log_weight"` `rvar` containing the weights. This leads to counterintuitive behavior, like the default output of the `rvar` (showing mean and sd) using unweighted versions of those statistics.

This PR addresses that issue in the following ways:

- It stores `rvar` weights as a `"log_weights"` attribute directly on the `rvar`, just like the `"nchains"` attribute is used to store chain count.
- `draws_rvars` no longer use a `".log_weight"` variable to store weights, instead storing them directly on each `rvar` they contain, and requiring all `rvar`s they contain to have the same weights (the same way it handles `"nchains"`).
- A new `log_weights()` function for `draws` and `rvar`s is added, which is a lower-level version of `weights(x, log = TRUE, normalize = FALSE)` that just returns the log weights stored in the object without transformation. I initially did not have this, but found it greatly eased programming with weights.
- `weight_draws(x, NULL)` is now allowed as the canonical way to remove weights from a `draws` object, since `remove_variables(x, ".log_weight")` does not work on `draws_rvars` objects anymore.
- All summary functions for `rvar`s have been updated to incorporate weights (with a couple of exceptions I haven't gotten to yet, see  **TODOs and Questions** below).

### Demo

``` r
set.seed(1234)
x = rvar(rnorm(1000))
x
#> rvar<1000>[1] mean ± sd:
#> [1] -0.027 ± 1

w1 = rexp(1000)
x1 = weight_draws(x, w1)
x1
#> weighted rvar<1000>[1] mean ± sd:
#> [1] -0.00087 ± 1

w2 = rexp(1000)
x2 = weight_draws(x, w2)
x2
#> weighted rvar<1000>[1] mean ± sd:
#> [1] -0.003 ± 0.96
```

You can't combine two rvars with different weights:

```r
x1 + x2
#> Error: Random variables have different log weights and cannot be used together:
#> <dbl> 0.794199981930473, -1.61888922585584, 1.02558084358998, -0.657945687118312, 0.132635682996154 ...
#> <dbl> 0.661721670766407, -1.46589074644228, -1.39312536919089, 0.318133129307739, 0.66043235310858 ...
```

The check for equality of weights is done on the internal weights using `identical()`, which should be fast, especially in cases where the two weight vectors are actually pointers to the same vector in memory (in which case the comparison is constant time). This does mean the weights vectors must be exactly the same (no tolerance for floating point error), but I suspect in most cases when weighting happens the exact same weight vector is being applied to many objects. In any case, if someone did encounter this issue they could simply assign the log weights from object to the other.

If one rvar is weighted and another is not, the weights of the weighted rvar are inherited, which I believe covers the use case of (weighted draws from some model) + (unweighted draws, e.g. used to simulate predictions):

```r
x1 + rvar(rnorm(1000, 1))
#> weighted rvar<1000>[1] mean ± sd:
#> [1] 0.96 ± 1.4
```

If you install the dev version of {ggdist}:

```r
remotes::install_github("mjskay/ggdist")
```

It supports weighted rvars in all functions (densities, CDFs, quantiles, all interval types and all point summaries):

Without weights:

```r
library(ggplot2)
library(ggdist)

set.seed(1234)
x = rvar(rnorm(10000, c(1,5)))

ggplot() + stat_slabinterval(aes(xdist = x))
```
image

With weights:

```r
xw = weight_draws(x, rep(c(1,2), 5000))
ggplot() + stat_slabinterval(aes(xdist = xw))
```
image

Weights should work basically everywhere:

```r
ggplot() + 
  stat_slabinterval(
    aes(xdist = xw), 
    point_interval = mode_hdi, 
    density = "histogram", 
    breaks = 50
  )
```
image

### TODOs and Questions

Some TODOs and questions remain:

- [ ] I still have to implement `density(<rvar>)`, `cdf(<rvar>)`, and `quantile(<rvar>)` / `quantile2(<rvar>)`. The first two are straightforward. For weighted quantiles, I have an implementation in {ggdist} that I can port over, but I may want to update it first; some thoughts on weighted quantiles are [here](https://htmlpreview.github.io/?https://github.com/mjskay/uncertainty-examples/blob/master/weighted-quantiles.html) and feedback is welcome. (In fact, since writing that document my thinking has changed a bit---I originally thought the way I suggested implementing weighted quantiles in that document is an improvement on ggdist's current implementation, but after further investigation I might be leaning back towards how I did it in ggdist originally...).
- [ ] I don't know if any of the other functions in `R/convergence.R` should be modified for weighted rvars. @avehtari?
- [ ] Have I missed anything else?

Would love for folks to kick the tires. I think once this is in we could also start thinking about what a successor to `summarise_draws()` might look like that supports weights (and solves the various other open issues on `summarise_draws()`).

### Copyright and Licensing


avehtari:
> I don't know if any of the other functions in R/convergence.R should be modified for weighted rvars.

Currently, everything else than `pareto_` functions assume non-weighted MCMC. I have so far assumed that MCMC and weighting are independent of each other (there might be some less common algorithms that jointly sample parameter values and weights).
- In PSIS paper experiments, I computed separately ESS for MCMC and ESS for PSIS and combined them as ESS_MCMC*ESS_PSIS/S, which worked well for getting MCSE that matched RMSE (given khat<0.7)
- I did not try what would happen if in ESS for MCMC computation we would just replace autocorrelation computation with functions that would use weighted means and variances, and I have not checked what it would produce, but I guess that would be not what we want.
- It would be nice to have a flag stating if rvar does not have Markov dependency, and then ESS and MCSE would be based just on weights.
- If there are both (assumed) Markov dependency and weights, we could follow the approach presented in PSIS paper for `ess_` and `mcse_` functions
- If we assume MCMC and weighting are independent, then `rhat_` and `rstar` could do the MCMC convergence check without weights (until we are aware of an algorithm that would have the weighting inside the MCMC already)
- `pareto_` functions are checking the tail(s) of a given argument, and it has been used to check tails of raw weights/ratios (r or r(theta) in PSIS paper notation), function of a variable (h or h(theta)), or the product (hr). With the weight support, they could automatically make the diagnostics for r and hr (and if no weights then just h). Here I'm assuming that we almost always use self-normalization so that we need to check the normalization (E[r]) and the quantity of interest (E[hr])

Pinging @n-kall , too

n-kall:
> * If there are both (assumed) Markov dependency and weights, we could follow the approach presented in PSIS paper for `ess_` and `mcse_` functions
For reference: Equations 6 (MCSE) and 7 (ESS) in preprint v6


> * `pareto_` functions are checking the tail(s) of a given argument, and it has been used to check tails of raw weights/ratios (r or r(theta) in PSIS paper notation), function of a variable (h or h(theta)), or the product (hr). With the weight support, they could automatically make the diagnostics for r and hr (and if no weights then just h). Here I'm assuming that we almost always use self-normalization so that we need to check the normalization (E[r]) and the quantity of interest (E[hr])

Any thoughts on how the two sets of diagnostics should be presented in summarise_draws? 
Would it make sense to have separate e.g. `pareto_khat_quantity`, `pareto_khat_weights` columns?

n-kall:
I'm currently working on updating the `pareto_`, `ess_` and `mcse_` functions for weighted rvars in a [fork](https://github.com/n-kall/posterior/tree/rvar_weights)

mjskay:
Okay, I think this is ready for review, pending two things:

1. If we want to add an attribute distinguishing MC vs MCMC to this PR. I would suggest we wait and do that as a separate PR to address #239, as it will involve further discussion / thought.
2. If we want to merge this PR to master and then merge @n-kall's branch on weighted diagnostics to master, or if we want to merge @n-kall's branch into this PR and then merge to master.

n-kall:
Re: option 2, I'll still need some more time to finish up the weighted mcse and ess. So I think it's better to merge without waiting for me

paul-buerkner:
I am at a conference and then on vacation for the next two weeks. Can someone else review this PR?

R/rvar-.R
```diff
@@ -90,7 +150,11 @@
 #' x
 #'
 #' @export
-rvar <- function(x = double(), dim = NULL, dimnames = NULL, nchains = NULL, with_chains = FALSE) {
+rvar <- function(
+  x = double(), dim = NULL, dimnames = NULL,
+  nchains = NULL, with_chains = FALSE,
+  log_weights = NULL
```
n-kall, line 156: I think `log_weights` needs input checking to ensure it is a vector. Currently it is possible to provide `log_weights` as a matrix which then causes issues when computing e.g. `mean` of the weighted `rvar`

mjskay: Good point. 

This should be true of weights everywhere right? `new_rvar()` calls down to `validate_weights()` so then something like `assert_atomic_vector(weights)` should go there. I can add that.

n-kall:
Are there any remaining blockers for this PR @mjskay @paul-buerkner ? It would be really nice to get this merged and start on changes that would depend on this. Happy to help out if there is something specific

n-kall:
Merge conflicts should be manageable, I'll take a look

Issue #333, opened 2024-01-10
should remove_variables be exported?


n-kall:
I was recently looking for a way to remove unwanted variables from a draws object, and I found that there is a `remove_variables` function (mentioned by in #331)  but it does not seem to be exported. Is this intentional?

Here, the generic is not exported, but the individual methods appear to be:
https://github.com/stan-dev/posterior/blob/1210f4a42eeb6b942572097d60a8cc2fb8540adf/R/remove_variables.R#L1C1-L14C2

Perhaps another option would be to add an `exclude` parameter in `subset_draws` (e.g. `subset_draws(example_draws(), variable = "theta", exclude = TRUE)` would remove all theta variables)

paul-buerkner:
I think we can export the generic and thus the methods actually. However, then we would have to add all kinds of validity checks, which we currently don't have in place there. I like the idea of adding an `exclude` option to subset_draws.
This could internally call remove_variables then.

BTW: the fact that the remove_variables methods are exported just ensures that they are registered as S3 methods to the generic. They won't actually be exported this way.

n-kall:
I think an `exclude` option for `subset_draws` could be implemented with setdiff like this. If this seems ok, I can make a PR

```r
subset_draws.draws_array <- function(x, variable = NULL, iteration = NULL,
                                     chain = NULL, draw = NULL, regex = FALSE,
                                     unique = TRUE, exclude = FALSE, ...) {
  if (all_null(variable, iteration, chain, draw)) {
    return(x)
  }

  x <- repair_draws(x)
  variable <- check_existing_variables(variable, x, regex = regex)
  iteration <- check_iteration_ids(iteration, x, unique = unique)
  chain <- check_chain_ids(chain, x, unique = unique)
  draw <- check_draw_ids(draw, x, unique = unique)

 ########## code for exclude option
  if (exclude) {
    if (!is.null(variable)) {
      variable <- setdiff(variables(x), variable)
    }
    if (!is.null(iteration)) {
      iteration <- setdiff(iteration_ids(x), iteration)
    }
    if (!is.null(chain)) {
      chain <- setdiff(chain_ids(x), chain)
    }
    if (!is.null(draw)) {
      draw <- setdiff(draw_ids(x), draw)
    }
  }
 #############

  x <- prepare_subsetting(x, iteration, chain, draw)
  if (!is.null(draw)) {
    iteration <- draw
  }
  x <- .subset_draws(x, iteration, chain, variable, reserved = TRUE)
[...]

PR #336, opened 2024-01-11
Exclude draws

n-kall:
#### Summary

Adds an `exclude` option to `subset_draws` that will remove the matched draws. Addresses #333 

TODO:

~~Tests~~

#### Copyright and Licensing


n-kall:
I tried to change as little code as necessary to implement this. But I think this is now ready for review

R/subset_draws.R
```diff
@@ -55,6 +61,22 @@ subset_draws.draws_matrix <- function(x, variable = NULL, iteration = NULL,
   iteration <- check_iteration_ids(iteration, x, unique = unique)
   chain <- check_chain_ids(chain, x, unique = unique)
   draw <- check_draw_ids(draw, x, unique = unique)
+
+  if (exclude) {
```
paul-buerkner, line 65: please input check exclude via as_one_logical

paul-buerkner (changes requested):
Thank you for working on this. I have left a few rather small comments I think

R/subset_draws.R
n-kall: Thanks! Keep forgetting to do this. I now added input checking into the check_* functions, as is done with the the other arguments `regex` and `unique`. This does mean that the inputs are not checked if all `draw`, `iteration`, `variable`, `chain` are NULL though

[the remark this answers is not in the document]
n-kall: Good point, I changed this

n-kall:
@paul-buerkner I think I addressed your comments now, also added more tests

PR #342, opened 2024-01-30
extract_variable_array and the with_indices parameter for `variables()`

### LINKED ISSUE #208: should variables() behave differently for `draws_rvars` format?
When working on https://github.com/stan-dev/cmdstanr/issues/582 I was reminded that `variables()` and `nvariables()` behave differently for the `draws_rvars` format compared to all other formats. I don't think this difference (treating vector/matrix variables as a single variable vs. treating elements as separate variables) is documented anywhere (or maybe I missed it). Should I add this information to the `help("draws-index")` page where these functions are documented?

### LINKED ISSUE #340: Add a function to extract single draw of multidimensional variable
Currently `extract_variable` doesn't work for matrix/array variables

In the following `p1` is a Pathfinder object from CmdStanR and `z_1` is defined in Stan as `matrix[M_1, N_1] z_1;`
It took me some time to figure out how to extract an R array with the dimensions `M_1`, `N_1`
```
drop(draws_of(subset_draws(as_draws_rvars(p1$draws()), variable='z_1', draw=1)[[1]]))
```

It would be nice to have a function for this. The above code works also for scalars, vectors, and 1D arrays.

mjskay:
#### Summary

This PR:

* closes #208 by giving `variables()`, `variables<-()`, `set_variables()`, and `nvariables()` a `with_indices` argument, which determines whether variable names are retrieved/set with (`"x[1]"`, `"x[2]"` ...) or without (`"x"`) indices.
* closes #340 by adding an `extract_variable_array()` function to extract variables with indices into arrays of iterations x chains x any remaining dimensions. @avehtari does this do what you need?

It also makes a few other related changes:

* For types that support `factor` variables (`draws_df`, `draws_list`, and `draws_rvars`), `extract_variable()` and `extract_variable_matrix()` now return a `factor` / `ordered` if the variable requested is a `factor` or `ordered`.
* It factors out a bunch of internal code for manipulating variable indices into `R/variable-indices.R`. Most important are:
   * `split_variable_names()` and `split_indices()`, which I suggest as the "canonical" functions for (1) splitting a variable into a string representing base name and indices and (2) parsing those indices if needed. The code for `split_variable_names()` was repeated in several places in the code base (implemented slightly differently). I tried several implementations on large vectors and the one I put there should be fast. 
   * `variable_names()` and `variable_names<-()`, which are functions with a `with_indices` parameter that can manipulate character vectors of variable names. These use `split_variable_names()` under the hood, and are the basis of the implementation of `with_indices` in `variables()` and `variables<-()`
   * `flatten_indices()` and `flatten_array()`, which are used to turn arrays into vectors where indices are embedded into variable names (used when converting from draws_rvars to other formats that embed indices into variable names).
   
   Factoring these functions out means if we later add support for more specialized parsing of variable names, we only have to do that in one place. It also gets us a bit closer to implementations of index parsing we could consider exporting (e.g. for #61), which could perhaps be a `parse_variable_names()` function built on `split_variable_names()` + `split_indices()`.

* Moves the code for `variables()`, `variables<-()`, `set_variables()`, and `nvariables()` from `R/draws-index.R` into `R/variables.R` and gives them their own documentation pages (`R/draws-index.R` and the combined doc page for variables/iterations/chains/draws was getting bloated, especially with the extra args on variables functions).
* Minor refactor of `while_preserving_dims()` -> `copy_dims()` and `while_preserving_levels()` -> `copy_levels()`
* Minor fix for a change in how the package doc page is specified in the latest version of roxygen2.

#### Copyright and Licensing


R/variables.R
```diff
@@ -0,0 +1,323 @@
...
+    all_var_matched_ixs <- c(all_var_ixs[!is.na(scalar_input_ixs)],
+                             all_var_ixs[!is.na(vector_input_ixs)])
+    # select missed input variables
+    missing_vars_mask <- rep_len(TRUE, length(variables))
+    missing_vars_mask[input_ixs] <- FALSE
+    missing_variables <- variables[missing_vars_mask]
+    # select matched all_variables maintaining the input variables order
+    variables <- all_variables[all_var_matched_ixs[order(input_ixs, all_var_matched_ixs)]]
+  } else {
+    missing_variables <- setdiff(variables, all_variables)
```
mjskay, line 279: This line is not hit by test coverage (see [here](https://app.codecov.io/gh/stan-dev/posterior/pull/342/blob/R/variables.R#L279)), but as far as I can tell it cannot be, as `scalar_only` is always `FALSE` (I assume this is a parameter that was used by some old code?). 

I have not changed this function except to use the new `split_variable_names()` on line 263. It only appears as all new because I moved it here from `R/draws-index.R`.

paul-buerkner:
thank you! Please let me know once this PR is ready for review!

mjskay:
Should be ready now

R/variables.R
paul-buerkner: I don't remember scalar_only. The way the variables are named suggests that I have I not written it. Was it previously used in rvars perhaps? I am fine with removing it as long as you don't see any value in keeping it for future use in rvars or similar functionality.

paul-buerkner (approved):
The PR looks good to me. Thank you!!

@avehtari could you check if `as_variable_array` does what you would like to do? If you give your okay too, I think this is ready for merging.

R/variables.R
paul-buerkner: I see. Interesting. I don't have any strong opinions about it. @jgabry and @MansMeg you seem to have been involved in the discussion. Any thoughts on this?

avehtari:
> closes https://github.com/stan-dev/posterior/issues/340 by adding an extract_variable_array() function to extract variables with indices into arrays of iterations x chains x any remaining dimensions. @avehtari does this do what you need?

Yes, it's helpful as it works the same way for any draws object type, and I can write my example in the issue as
```
p1$draws() |> subset_draws(draw=1) |> extract_variable_array(variable='r_id') |> drop()
```
or if I want to get all draws array
```
p1$draws() |> merge_chains() |> extract_variable_array(variable='r_id') |> drop()
```

(I can also drop `drop()` and index the 4 dim array)

paul-buerkner:
Great! @mjskay anything to change from your side? Otherwise, feel free to merge (or tell me such that I merge).

paul-buerkner:
Agreed.

avehtari:
I'm fine with `with_indices` and can't come up with better suggestions. I tested using `variables()`, `nvariables()`, and `set_variables()`, and having `with_indices` parameter is awesome!

mjskay:
Great, thanks @avehtari! I will take that as a strong endorsement and merge :)

PR #344, opened 2024-03-01
update psis paper reference

avehtari:
psis paper has been accepted for publication


paul-buerkner:
Congratulations! I almost cannot believe they eventually accepted this paper after having it in review for about 6(?) years :-D

PR #354, opened 2024-03-13
Add relative efficiency function

### LINKED ISSUE #280: [Feature] Add relative_eff function
For some diagnostics, such as pareto-k diagnostics, we may want to calculate the relative efficiency (r_eff). The `loo` package has an implementation of this `loo::relative_eff`. The documentation states that it 'computes the MCMC effective sample size divided by the total sample size', but the output is not equal to `ess_bulk(x)/ndraws(x)` nor `ess_basic(x)/ndraws(x)`, so perhaps the method in `loo` could be adapted and moved to posterior?

n-kall:
#### Summary

This would fix #280. 

I am revisiting r_eff while implementing the weighted MCSE and ESS, which should take this into account (if draws are from MCMC).

I implemented `relative_eff` as `ess_basic(x) / ndraws(x)`. This gives slightly lower (about 1%) results than `loo::relative_eff`. I guess due to the difference between `loo:::rfun_ess` and `posterior::ess_basic`

#### Copyright and Licensing


n-kall:
I changed this to a draft as I am now thinking about whether a single `relative_eff` function is the way to go. Perhaps there should be one for different quantities (like ESS) or none at all

n-kall:
As a single `r_eff` function doesn't really make sense (it should be dependent on the quantity of interest), I will close this PR.

PR #356, opened 2024-03-15
modify summarise_draws internals not to coerce values to same type

### LINKED ISSUE #355: In summarise draws, a function that returns a character changes all columns to character
If a summary function that returns a character is used in `summarise_draws`, it makes all columns `<chr>` which messes up formatting.

```r
example_draws() |>
   summarise_draws(x, char_fun = \(x) "word", mean)
```
Returns
```
# A tibble: 10 × 3
   variable char_fun      mean
   <chr>    <chr>         <chr>
 1 mu       word          Inf
 2 tau      word          4.16356885610418
 3 theta[1] word          6.74893947963962
 4 theta[2] word          5.25331634990726
 5 theta[3] word          3.04393475572924
 6 theta[4] word          4.85842854299997
 7 theta[5] word          3.22258991784476
 8 theta[6] word          3.98696993634695
 9 theta[7] word          6.5030995214442
10 theta[8] word          4.56520199862817

```

This could be fixed by using `type.convert(out, as.is = TRUE)` on the output tibble in `summarise_draws_helper`

n-kall:
#### Summary

This would fix #355.

As it is affecting `summarise_draws`, which is one of the main functions in `posterior`, I think this change needs to be considered carefully. It might be less efficient than currently implemented, so I am opening this PR now to check how the benchmarks for `summarise_draws` perform.

#### Copyright and Licensing

### DIFF
--- a/R/misc.R
+++ b/R/misc.R
@@ -14,6 +14,31 @@ named_list <- function(names, values = NULL) {
   setNames(values, names)
 }
 
+
+# flatten a list that may contain vectors into a single level list
+unnest <- function(x) {
+
+  out <- list()
+
+  for (i in seq_along(x)) {
+    name_i <- names(x[i])
+    if (length(x[[i]]) > 1) {
+      if (rlang::is_named(x[[i]])) {
+        name_j <- names(x[[i]])
+      } else {
+        name_j <- paste0(name_i, ".", c(1:length(x[[i]])))
+      }
+      for (j in seq_along(x[[i]])) {
+        out[[name_j[j]]] <- x[[i]][[j]]
+      }
+    } else {
+      out[[name_i]] <- x[[i]]
+    }
+  }
+  out
+}
+
+
 # unlist lapply output
 ulapply <- function(X, FUN, ..., recursive = TRUE, use.names = TRUE) {
   unlist(lapply(X, FUN, ...), recursive, use.names)
--- a/R/summarise_draws.R
+++ b/R/summarise_draws.R
@@ -341,30 +341,21 @@ summarise_draws_helper <- function(x, funs, .args) {
   variables_x <- variables(x)
   # get length and output names, calculated on the first variable
   out_1 <- create_summary_list(x, variables_x[1], funs, .args)
-  the_names <- vector(mode = "list", length = length(funs))
-  for (i in seq_along(out_1)){
-    if (rlang::is_named(out_1[[i]])) {
-      the_names[[i]] <- names(out_1[[i]])
-    } else if (length(out_1[[i]]) > 1) {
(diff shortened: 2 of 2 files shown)


n-kall:
Looks like a >1% increase to `summarise_draws` times, which I don't think is acceptable

paul-buerkner:
How much increase exactly? 1% would barely matter I think

n-kall:
> * ❗🐌summarise_draws_100_variables: 706ms -> 715ms [+0.92%, +1.51%]
> 
> * ❗🐌summarise_draws_10_variables: 105ms -> 107ms [+0.93%, +2.1%]

Based on these ^

n-kall:
I've now added tests checking that the column types are as expected, and if the speed is ok, this is ready for review

mjskay:
FWIW, I think the unnesting / flattening might be doable as a cbind on row vectors (or maybe vec_cbind / vec_rbind for safety) to create a 1-row data frame. Dunno if that helps at all.

paul-buerkner:
Speed is okay. Still, is this already the best way to implement the flattening?

n-kall:
> unnesting / flattening might be doable as a cbind on row vectors (or maybe vec_cbind / vec_rbind for safety) to create a 1-row data frame.

Thanks for the tip! Simplifying this unnesting/flattening would, I think, be beneficial. However, I can't quite get the right structure with `rbind` or `vec_rbind` as the `quantile2` column ends up with vectors of length 2 (as opposed to separate columns as in `unlist`). Looks like it would still need some kind of column splitting (e.g. `tidyr::unnest_wider`)

```r
create_summary_list(example_draws(), 3, list(rhat = rhat_basic, quantile2 = quantile2), NULL) |>
   vec_rbind()

##       rhat            quantile2
## 1 1.014967 -1.231353, 18.874003

create_summary_list(example_draws(), 3, list(rhat = rhat_basic, quantile2 = quantile2), NULL) |>
   unlist()

##         rhat  quantile2.q5 quantile2.q95
##     1.014967     -1.231353     18.874003
```

mjskay:
Ah right... you might be able to use cbind if you conditionally transpose (or rbind) vectors of length > 1 to turn them into row vectors (not on a computer atm so I can't try it)

n-kall:
> is this already the best way to implement the flattening

I don't think the issue is so critical that the fix is needed right away, so perhaps it's better to leave this open in case some other ideas come up (e.g. if @mjskay's method works out). However, I won't spend more time on this now

n-kall:
@paul-buerkner as discussed, I modified this PR to use `purrr::flatten`. This does have an issue where a function that returns an unnamed list or vector won't be included in the output, but there is a test for that case (failing).

Issue #346, opened 2024-03-04
pareto diagnostics return named values, preventing renaming


n-kall:
Currently the `pareto_*` summary functions return named vectors without the preceding `pareto_`. e.g. `pareto_convergence_rate` returns a numeric named `convergence_rate`.

This was done as the function names were deemed too long for column names. However, this also has the side effect that the columns can not be renamed in `summarise_draws` (and `convergence_rate` is still quite long). I think all other built-in diagnostics that return single values are unnamed, so this inconsistency is a bit unexpected.

The functions names and lengths are:
`pareto_convergence_rate` (23 characters, still 16 without `pareto_`)
`pareto_khat_threshold` (21 characters, 14 without `pareto_`)
`pareto_min_ss` (13 characters, 6 without `pareto`)
`pareto_khat`. (11 characters, 4 without `pareto_`)

In contrast, the longest name for other built-in summary functions is `rhat_nested` (11 characters, also a recent addition), and others are shorter than 10 characters (e.g. `ess_tail`).

I see two options to fix this:

1. shorten the function names where possible and then change to return unnamed values
2. keep returning named values, but shorten the names for `convergence_rate` and `khat_threshold` (maybe `convrate` and `khat_thresh`?)

PR #370, opened 2024-05-16
Change pareto functions to return unnamed numerics

### LINKED ISSUE #359: pareto_khat should always return the same data structure
That's unfortunate. @avehtari this seems to be a consequence of switching to using some of the functions from posterior. I guess we missed this difference and didn't have a test to detect it. I think `posterior::pareto_khat` should always return a list instead of having different behavior depending on if the result is NA (so it should just return `list(NA)`), but that would now break backwards compatibility in posterior if we change that. I guess we can just work around this in loo. I'll submit a PR.

_Originally posted by @jgabry in https://github.com/stan-dev/loo/issues/263#issuecomment-2025875767_

n-kall:
#### Summary

This changes `pareto_*` functions to no longer return named numeric values if there is just a single value output.
This applies to `pareto_khat`, `pareto_khat_threshold`, `pareto_min_ss`, and `pareto_convergence_rate`.

`pareto_diags` returns a named vector of all of these. And `pareto_smooth` is not affected by this.

This also changes the return type of `pareto_khat` to be an unnamed numeric rather than a list, which would require a change in loo (specifically [E_loo](https://github.com/stan-dev/loo/blob/1a4cfce4fdb79f04c4c79ba4411f6056363a8ff3/R/E_loo.R#L293)). As there is already an issue that needed to be worked around (stan-dev/loo/issues/263) I think this is worth addressing now.

This would close #359 and #346 

#### Copyright and Licensing


paul-buerkner:
Looks good to me. Do you want to make any changes before I merge?

n-kall:
I think it's ready

paul-buerkner:
great! Merging now. Thank you!

Issue #386, opened 2024-12-13
Most recent commit breaks CmdStanR models with 1 chain


jgabry:
@paul-buerkner I think this commit https://github.com/stan-dev/posterior/commit/79d4521b943e44f4ac31636c4488d9e2cfeac3ec is causing problems for CmdStanR models with 1 chain. The reason is that with that commit a list of a single data frame now gets passed to `posterior::: as_array_matrix_list` because `is_matrix_list_like` is TRUE. This results in an error because of these lines in `as_array_matrix_list`: https://github.com/stan-dev/posterior/blob/79d4521b943e44f4ac31636c4488d9e2cfeac3ec/R/as_draws_array.R#L232-L234 

If the list has two data frames there's no error, which is why it only affects runs of a single chain. 

This can be demonstrated without CmdStanR with a toy example. Here I first check the CRAN version (no error), then the latest posterior on GitHub (error), then install from posterior on GitHub from the commit _before_ the latest commit (no error): 

```r
remotes::install_cran("posterior")
test <- list(
  data.frame(x = rnorm(100), y = rnorm(100))
)
posterior::as_draws_array(test) # no error

# need to restart R
remotes::install_github("stan-dev/posterior") # includes latest commit
test <- list(
  data.frame(x = rnorm(100), y = rnorm(100))
)
posterior::as_draws_array(test) # error

# need to restart R
remotes::install_github("stan-dev/posterior", ref = "d0f8b76") # this commit is from the PR before your latest commit
test <- list(
  data.frame(x = rnorm(100), y = rnorm(100))
)
posterior::as_draws_array(test) # no error 
```

When a CmdStanR model has a single chain this error happens when `as_draws_array()` is called on the list of a single data frame that is created when reading in the single CSV file.

jgabry:
I believe this is the source of the error in https://github.com/stan-dev/cmdstanr/issues/1050. It took me a while to notice that this it was using the latest commit of posterior.

jgabry:
@tillahoffmann also confirmed in https://github.com/stan-dev/cmdstanr/issues/1050#issuecomment-2542403779 that he can replicate this posterior behavior, so I'm pretty sure this is the issue causing the errors in CmdStanR.

paul-buerkner:
Thank you for finding this error. I am sorry I didn't respond earlier. I didn't get email notifications from github. Do you have a suggestion for how to fix this without having to revert to the commit? I would like to keep this feature if we can make it work safely with cmdStanR

[...]

PR #387, opened 2024-12-17
 Convert lists of matrices to `draws_array` objects

paul-buerkner:
#### Summary

This PR allows to convert lists of matrices to `draws_array` objects. We had to initially revert as it was causing issues in cmdstanr models with one chain (#386). 

In this PR I also fix the issue with `as_array_matrix_list` such that we now should no longer see any issues. Interestingly, I believe that the bug in `as_array_matrix_list` existed also previously (for a long time I think). Just nobody used `as_array_matrix_list` until I made more objects to trigger it being called. Anyway it works now. 

@jgabry can you perhaps quickly check and confirm that it now works as expected? It passes all posterior and cmdstanr tests for me and works with one-chain models.

#### Copyright and Licensing

### DIFF
--- a/R/as_draws_array.R
+++ b/R/as_draws_array.R
@@ -227,15 +227,8 @@ variance.draws_array <- function(x, ...) {
 # convert a list of matrices to an array
 as_array_matrix_list <- function(x) {
   stopifnot(is.list(x))
-  if (length(x) == 1) {
-    tmp <- dimnames(x[[1]])
-    x <- x[[1]]
-    dim(x) <- c(dim(x), 1)
-    dimnames(x) <- tmp
-  } else {
-    x <- abind::abind(x, along = 3L)
-  }
-  x <- aperm(x, c(1, 3, 2))
+  x <- abind::abind(x, along = 3L)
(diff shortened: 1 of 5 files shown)


R/as_draws_array.R
```diff
@@ -216,15 +216,8 @@ variance.draws_array <- function(x, ...) {
...
-  if (length(x) == 1) {
-    tmp <- dimnames(x[[1]])
-    x <- x[[1]]
-    dim(x) <- c(dim(x), 1)
-    dimnames(x) <- tmp
-  } else {
-    x <- abind::abind(x, along = 3L)
-  }
-  x <- aperm(x, c(1, 3, 2))
+  x <- abind::abind(x, along = 3L)
```
jgabry, line 219: great, this should avoid the cmdstanr error with 1 chain

jgabry (approved):
I accidentally broke it when fixing a merge conflict, but should be good now. Your changes look good. Will merge when it finishes running the tests.

jgabry:
Will go ahead and merge. The failure on ubuntu-latest (3.6) is unrelated.

PR #266, opened 2023-01-02
minor change in ESS estimation in mcse_sd based on more simulations

avehtari:
#### Summary

mcse_sd was estimating ESS from abs(x-E[x]). In my old simulations, this had worked better than (x-E[x])^2. It's likely that the change in how the rest of MCSE for sd is estimated (PR #233) did affect that now the more logical (x-E[x])^2 works better, as demonstrated by simulations run by @sethaxen.

Here are the plots for the new simulation results
image
image

#### Copyright and Licensing

By submitting this pull request, the copyright holder is agreeing to 
license the submitted work under the following licenses:

- Code: BSD 3-clause (https://opensource.org/licenses/BSD-3-Clause)
- Documentation: CC-BY 4.0 (https://creativecommons.org/licenses/by/4.0/)

PR #388, opened 2024-12-20
use the same equation in ess_sd as in mcse_sd

avehtari:
@sethaxen pointed out that PR #266 had made the change only in `mcse_sd()`, but based on the same justification, the change should have been made to `ess_sd()`, too. This makes the change for `ess_sd()` and updates the test value range for `ess_sd()`.


paul-buerkner:
Thank you!

PR #390, opened 2025-01-29
fix documentation of u_scale()

avehtari:
u_scale() function had quite old documentation text, and this PR fixes it to match with the actual implementation


paul-buerkner:
Thanks!

Issue #338, opened 2024-01-19
Add functions for PIT and LOO-PIT


TeemuSailynoja:
`loo_pit` is currently in `rstantools` but needs implementation for discrete variables, and should probably be moved to `posterior`.
`pit` isn't currently available, but is used, for example, by `bayesplot`, where I have currently implemented the computation.

My questions are:
1 . PIT would require observations in addition to the draws object, which to my understanding isn't the case for other functions. Is this a problem?
2. Should there be a new subtitle on the functions page for PIT and LOO-PIT to fall under?
3. Should they be in the same function as regular PIT is just LOO-PIT with even weights?

paul-buerkner:
I think that PIT in posterior does make sense. Will we want to use it for other stuff than plotting though? If yes, than posterior is I think good. If we only ever want to use it for bayesplot, then perhaps also bayesplot could be a reasonable target(?)

paul-buerkner:
Just to clarify, are we talking about the methods or the generic (or both)?

avehtari:
- PIT values could be summarised with scalar diagnostic value, but the preferred way is graphics, but we may want to use other plots than what is available in bayesplot (which is what I was doing)
- 1. methods, 2. it would be nice if the method function could be called with any posterior draws object type

paul-buerkner:
sounds like good arguments for posterior indeed.

Aki Vehtari ***@***.***> schrieb am Fr., 19. Jan. 2024, 18:55:

jgabry:
Sounds good to me too

avehtari:
@TeemuSailynoja can this issue be closed?

florence-bockting:
This Issue has been addressed in #395 ([link to Issue](https://github.com/stan-dev/posterior/pull/395#issuecomment-2703510722))

PR #395, opened 2025-03-06
Add pit functions

TeemuSailynoja:
#### Summary

Add functions for computing probability integral transforms. This includes both the regular PIT of observations with respect to posterior predictive draws, as well as, a weighted version, where LOO weights can optionally be supplied. 

#### Copyright and Licensing


TeemuSailynoja:
This would close #338.

paul-buerkner:
Thank you! Are there tests in place already?

TeemuSailynoja:
No, this is still a draft PR as I only have the default method and have not rested with rvars or draws objects.

TeemuSailynoja:
@paul-buerkner I think this should now be at a state ready for a review.

R/pit.R
```diff
@@ -0,0 +1,146 @@
...
+
+# validate weights and return log weights
+validate_loo_weights <- function(loo_weights, draws, log = TRUE) {
+  checkmate::assert_numeric(loo_weights, any.missing = FALSE)
+  checkmate::assert_flag(log)
+  if (!all(dim(loo_weights) == dim(draws))) {
+    stop_no_call("Dimension of `loo_weights` must match that of `x`.")
+  }
+  if (!all(is.finite(loo_weights))) {
+    stop_no_call("All weigths in `loo_weights` must be finite.")
```
n-kall, line 133:
```suggestion
    stop_no_call("All weights in `loo_weights` must be finite.")
```

R/pit.R
```diff
@@ -0,0 +1,146 @@
...
+    if (any(loo_weights < 0)) {
+      stop_no_call("`loo-weights` must be non-negative when log = FALSE.")
+    }
+    loo_weights <- log(loo_weights)
+  }
+  loo_weights
+}
+
+normalize_log_weights <- function(log_weights) {
+  apply(log_weights, 2, \(col) col - log_sum_exp(col))
```
n-kall, line 145: I think the `\(x)` syntax requires R 4.1.0

R/pit.R
TeemuSailynoja: Oh no! Not this again!

R/pit.R
TeemuSailynoja: Good catch! Fixed that.

R/pit.R
```diff
@@ -0,0 +1,146 @@
...
+        y,
+        as_draws_matrix(c(loo_weights)),
+        log
+      ),
+      dim(x),
+      dimnames(x)
+    )
+  }
+
+  pit
```
paul-buerkner, line 120: what kind of object does pit.rvar return? doc says a vector of length = length(y). Is this also the case here?

paul-buerkner: Perhaps more generally, can you provide more context what the pit.rvar implementation does differently than the default/draws_matrix one?

R/pit.R
```diff
@@ -0,0 +1,146 @@
...
+#'   assessment for count data.  Biometrics 65(4), 1254–1261 (2009).)
+#'
+#' @return A numeric vector of length `length(y)` containing the PIT values.
+#'
+#' @export
+pit <- function(x, y, loo_weights = NULL, log = TRUE) UseMethod("pit")
+
+#' @rdname pit
+#' @export
+pit.default <- function(x, y, loo_weights = NULL, log = TRUE) {
```
paul-buerkner, line 32: Are weights only valid if they come from loo? Or can they be, in theory, any kind of weights? If so, can we just call them weights?

R/pit.R
```diff
@@ -0,0 +1,146 @@
...
+#'
+#' The probability integral transform (PIT). LOO-PIT is achieved with weighted sample.
+#'
+#' @param x (draws) A [`draws_matrix`] object or one coercible to a
+#' `draws_matrix` object.
+
+#' @param y (observations) A 1D vector of observations. Each element of `y`
+#' corresponds to a column in `x`.
+
+#' @param loo_weights A [`draws_matrix`] object or one coercible to a
```
paul-buerkner, line 11: highlight that by default these are expected to be log weights? What is more, do is log = TRUE a good default? Can we assume users to see this?

R/pit.R
```diff
@@ -0,0 +1,146 @@
...
+#' `draws_matrix` object. `loo_weights`` should have the same dimensions as `x`.
+
+#' @param log Are the weights provided in `loo_weights` log-weights.
+
+#' @details The `pit()` function computes the probability integral transform of
```
paul-buerkner, line 16: could pit be explained with one equation so people don't have to google it?

R/pit.R
paul-buerkner, line 32: for extensibility, please add ... to all methods and the generic

R/pit.R
```diff
@@ -0,0 +1,146 @@
...
+#    `loo_weights`, which enables for example the computation of LOO-PITs.
+#'
+#'   If `y` and `x` are discrete, randomisation is used to obtain continuous PIT
+#'   values. (see, e.g., Czado, C., Gneiting, T., Held, L.: Predictive model
+#'   assessment for count data.  Biometrics 65(4), 1254–1261 (2009).)
+#'
+#' @return A numeric vector of length `length(y)` containing the PIT values.
+#'
+#' @export
+pit <- function(x, y, loo_weights = NULL, log = TRUE) UseMethod("pit")
```
paul-buerkner, line 28: this signature is very restrictive. I would suggest the signature (x, y, ...) for better extensibility.

R/pit.R
```diff
@@ -0,0 +1,146 @@
...
+  if (!is.null(loo_weights)) {
+    loo_weights <- as_draws_matrix(loo_weights)
+  }
+  pit(x, y, loo_weights, log)
+}
+
+#' @rdname pit
+#' @export
+pit.draws_matrix <- function(x, y, loo_weights = NULL, log = TRUE) {
+  if (length(y) != ncol(x)) {
```
paul-buerkner, line 43: Could y be better validated? E.g., check this it is numeric or something along these lines?

R/pit.R
```diff
@@ -0,0 +1,146 @@
...
+        pit_sup <- pit + exp(log_sum_exp(loo_weights[sel_sup, j]))
+      }
+
+      pit <- runif(1, pit, pit_sup)
+    }
+    pit
+  }, FUN.VALUE = 1.0)
+
+  if (any(pit > 1)) {
+    warning_no_call(
```
paul-buerkner, line 79: Do we want to warn about this or just set it to 1 without warning?

R/pit.R
```diff
@@ -0,0 +1,146 @@
...
+  }
+  if (is.null(loo_weights)) {
+    pit <- array(
+      runif(length(y), Pr(x < y), Pr(x <= y)),
+      dim(x),
+      dimnames(x)
+    )
+  } else {
+    pit <- array(
+      pit(
```
paul-buerkner, line 109: Perhaps name the arguments to this call here for additional safety if we change something in the future?

paul-buerkner (changes requested):
Thank you! Looks good overall. I have noted down a couple of questions and suggestions.

R/pit.R
TeemuSailynoja: I'll add a description of this to the documentation. Also, I'll add an example to the function. For `rvar` the implemented method keeps the dimensions of the rvar and the observation in `y`, which are checked to match.

R/pit.R
TeemuSailynoja: I have changed them to weights. The difference to the weights already available to draws objects in `.weights` is, that this function requires a weight for each draw and each variable.
We were discussing with @n-kall that in the presence of the per-draw `.weights` they should also be taken into account.

paul-buerkner: Okay. Makes sense.

R/pit.R
paul-buerkner: thanks.

R/pit.R
avehtari: There should be a tolerance, as I've seen `pit > 1` due to floating point arithmetic inaccuracy. Values within tolerance should be set to 1.

R/pit.R
TeemuSailynoja: This could be a future extension for the functionality of `pit()`.

R/pit.R
TeemuSailynoja: I added a function for `validate_y`

```r

validate_y <- function(y, x = NULL) {
  if (!is.numeric(y)) {
    stop_no_call("`y` must be numeric.")
  }
  if (anyNA(y)) {
    stop_no_call("NAs not allowed in `y`.")
  }
  if (!is.array.rvar(x)) {
    if (!is.vector(y, mode = "numeric") || length(y) != nvariables(x)) {
      stop_no_call("`y` must be a vector of length `nvariables(x)`.")
    }
  } else if (!is.null(x)) {
    if (length(x) != length(y) || any(dim(y) != dim(x))) {
      stop_no_call("`dim(y)` must match `dim(x)`.")
    }
  }
  y
}
```

Currently, there is no other function in `posterior` that would use this validation, and the requirement of matching dimensions with `x` seems specific to this `pit()`.
Just in case of future extensions, I left `x` as an optional argument.

R/pit.R
TeemuSailynoja: In the latest commit, I added a tolerance for this (1e-10). I was no longer able to create a case, where this tolerance would be exceeded, so I removed the unit test for that, but kept the check and warning message.

R/pit.R
TeemuSailynoja: I changed the default to FALSE, and used the same doc-string as `weight_draws` to highlight the default value.

R/pit.R
TeemuSailynoja:
```r
#' @details The `pit()` function computes the probability integral transform of
#'   `y` using the empirical cumulative distribution computed from the samples
#'   in `x`. For continuous valued `y` and `x`, the PIT for the elements of `y`
#'   is computed as the empirical cumulative distribution value:
#'
#'     PIT(y_i) = Pr(x_i < y_i),
#'
#'   where x_i, is the corresponding set of draws in `x`. For `draws` objects,
#'   this corresponds to the draws of the *i*th variable, and for `rvar`
#'   the elements of `y` and `x` are matched.
#'
#'   The draws in `x` can further be provided (log-)weights in
#    `weights`, which enables for example the computation of LOO-PITs.
#'
#'   If `y` and `x` are discrete, randomisation is used to obtain continuous PIT
#'   values. (see, e.g., Czado, C., Gneiting, T., Held, L.: Predictive model
#'   assessment for count data.  Biometrics 65(4), 1254–1261 (2009).)

```

R/pit.R
TeemuSailynoja: added.

R/pit.R
TeemuSailynoja: I changed the default signature to the suggested one.

PR #398, opened 2025-03-20
Update workflows

TeemuSailynoja:
#### Summary
This should fix two of the failing workflow checks.

In short, `touchstone-receive.yaml` and `touchstone-comment.yaml` use an outdated action specified in `lorenzwalthert/touchstone/actions/comment@v1`. Although there is no new version release, the repo has been updated and the checks pass with the latest commit ID.

#### Copyright and Licensing

### DIFF
--- a/.github/workflows/touchstone-comment.yaml
+++ b/.github/workflows/touchstone-comment.yaml
@@ -16,6 +16,6 @@ jobs:
     if: >
       ${{ github.event.workflow_run.event == 'pull_request' }}
     steps:
-      - uses: lorenzwalthert/touchstone/actions/comment@v1
+      - uses: lorenzwalthert/touchstone/actions/comment@a02c45d24459777f4bbed5e5cc0c94571799f51f
         with: 
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}


TeemuSailynoja:
The last two seem to be failing because of missing system packages.

paul-buerkner:
Thanks! Do the last two fail because of some pandoc package not being available? If so, I had the same issue in brms and fixed it by fixing the ubuntu version (https://github.com/paul-buerkner/brms/blob/master/.github/workflows/R-CMD-check.yaml). Do you think it could help here too? Not sure what to do with the mac one though.

TeemuSailynoja:
Missing system packages:
Ubuntu - curl (available for Ubuntu from `libcurl4-openssl-dev`)
macOS is missing the file `libintl.h`, which apparently comes with `brew install gettext`

TeemuSailynoja:
The ubuntu-latest is trying to run the tests on Ubuntu 24.04.2. I would expect fixing the version to Ubuntu 22.04 to do the trick. The workflow file is actually calling 
```remotes::system_requirements("ubuntu", "18.04")``` [here](https://github.com/stan-dev/posterior/actions/runs/13965672474/workflow?pr=398#L73). remotes does not seem to include 24.04 yet.

TeemuSailynoja:
For some reason, the workflow file for touchstone comment from the master branch is being used instead of the updated one in the PR. Atleast that is what I see when I inspect the run: https://github.com/stan-dev/posterior/actions/runs/13966245808

PR #400, opened 2025-03-27
Update log_sum_exp

### LINKED ISSUE #399: Corner cases of log_sum_exp
I ran into a few surprising behaviours with the [log_sum_exp-function](https://github.com/stan-dev/posterior/blob/4e1cc8ad10507607c73b3bd71a3e968f956a98fa/R/misc.R#L211C1-L215C2) and wanted to ask, if these are intended, or if I should add some changes to the function in #395?

```r
log_sum_exp <- function(x) {
  max <- max(x)
  sum <- sum(exp(x - max))
  max + log(sum)
}
```

### 1. x is empty
```r
r$> posterior:::log_sum_exp(numeric(0))
[1] -Inf
Warning message:
In max(x) : no non-missing arguments to max; returning -Inf

r$> sum(numeric(0))
[1] 0
```
Using `max(x, warning = FALSE)` removes the warning and returns -Inf, but I can't find documentation for what else could throw a warning from `max()`.
**Should we suppress the warning, or keep the current behaviour?**

### 2. x contains only negative infinites
Here, I would expect this to behave like summing zeros, that is, I would expect
`log_sum_exp(-Inf) == -Inf`,
instead, we get 
```r
r$> posterior:::log_sum_exp(-Inf)
[1] NaN
```
as the second row of the function contains -Inf + Inf, and produces a NaN.
**This I think should be changed.**

### 3. x contains Inf
Here, I would expect to have `log_sum_exp(Inf) == Inf`, but instead have
```r
r$> posterior:::log_sum_exp(Inf)
[1] NaN
```
again, this is caused by the step computing `Inf - Inf`, which results in `NaN`.

## Suggestion
Change `log_sum_exp` to
```r
log_sum_exp <- function(x) {
  max <- max(x, warnings = FALSE)
  if (max == Inf) {
    res <- Inf
  } else if (max == -Inf) {
    res <- 0
  } else {
    sum <- sum(exp(x - max))
    res <- max + log(sum)
  }
  res
}
```

TeemuSailynoja:
## Summary

Add behaviour to the function in the corner cases of infinite values or empty input.
Fixes #399 

#### Copyright and Licensing

### DIFF
--- a/R/misc.R
+++ b/R/misc.R
@@ -1,4 +1,4 @@
-# initialize a named list
+  # initialize a named list
 # @param names names of the elements
 # @param values optional values of the elements
 named_list <- function(names, values = NULL) {
@@ -209,9 +209,16 @@ escape_all <- function(x) {
 
 # numerically stable version of log(sum(exp(x)))
 log_sum_exp <- function(x) {
-  max <- max(x)
-  sum <- sum(exp(x - max))
-  max + log(sum)
+  max <- max(as.numeric(x), warnings = FALSE)
+  if (max == -Inf) {
+    res <- 0
+  } else if (max == Inf) {
+    res <- Inf
+  } else {
+    sum <- sum(exp(x - max))
+    res <- max + log(sum)
+  }
+  res
 }
 
 # simple version of destructuring assignment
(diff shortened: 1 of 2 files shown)


paul-buerkner:
Thanks! Would you mind adding some tests that ensure that all the cases of confusing behavior you saw previously are now fixed?

paul-buerkner:
Thank you!

PR #413, opened 2025-10-17
make cutoff always smaller than the tail values

### LINKED ISSUE #411: gpdfit should not be called with zeros
The cutoff value is calculated as the value smaller than the smallest tail draw. If this value is the same as the first tail draw, zeroes can result, which cannot be fit with gpd

avehtari:
fixes #411


n-kall:
Looks good

PR #414, opened 2025-10-17
return NA if xstar == x[1]

### LINKED ISSUE #412: gpdfit should check if first quartile is equal to minimum, and then return NA
gpdfit should return `NA` if `xstar == x[1]`

avehtari:
fixes #412


paul-buerkner:
Could you add tests as part of this PR?

avehtari:
I added some tests. Suggestions for more tests?

paul-buerkner:
Looks good, merging now.

PR #416, opened 2025-10-19
Check ndraws_tail

### LINKED ISSUE #410: pareto_smooth should check that ndraws_tail is smaller than ndraws
Currently, `ndraws_tail` can be larger or the same the length of `x`, which causes NA or Inf values. This should be checked on input

VisruthSK:
Fixes #410.

### DIFF
--- a/R/pareto_smooth.R
+++ b/R/pareto_smooth.R
@@ -306,7 +306,7 @@ pareto_smooth.default <- function(x,
   }
 
   if (ndraws_tail < 5) {
-    warning("Number of tail draws cannot be less than 5. ",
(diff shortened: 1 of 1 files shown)


n-kall:
Looks good to me, thanks for the fix

paul-buerkner:
Is there a reason one is `warning` and one is `warning_no_call`? both should be the latter right?

VisruthSK:
I had just moved the ndraws_tail < 5 check out of the `tail = "both"` branch so it is always checked--the tail size check I wrote myself and copied the style of other checks I saw. If you think the <5 check should also be using `warning_no_call()` I can change it.

paul-buerkner:
Thank you!

PR #417, opened 2025-10-24
New pkgdown theme

VisruthSK:
@jgabry and I have been updating the Stan R package themes to better match the main Stan page. This PR uses the new theme & use a GitHub Action to automatically build the dev version of the site on PRs and the release version on releases.

This PR moves the site build into the gh-pages branch, so GH Pages will simply need to point to the correct branch. The move is so that dev website builds on every push to main don't clutter the commit log.

### DIFF
--- a/.github/workflows/pkgdown.yaml
+++ b/.github/workflows/pkgdown.yaml
@@ -52,4 +52,4 @@ jobs:
         with:
           clean: false
           branch: gh-pages
-          folder: "."
+          folder: "docs"


paul-buerkner:
Thank you! I don't feel comfortable reviewing this one unless necessary. @jgabry can you take a look?

jgabry:
> This PR moves the site build into the gh-pages branch, so GH Pages will simply need to point to the correct branch. The move is so that dev website builds on every push to main don't clutter the commit log.

So basically after merging this (or before?) I would just need to go into the repository settings and chance this to gh-pages instead of master/docs, right? Also did we decide to use gh-pages/docs or just root? I'm forgetting right now but I know we talked about this.

VisruthSK:
I'm also not sure which we decided on. Should be very easy to swap, but we should probably pick something and keep that consistent.

I swapped it for a few minutes last night^ and everything seemed ok, so I'd say maybe swap, poke about a bit, fix anything then merge?

^speaking of, the R packages role seems to get maintainer privileges to this repo, not sure if that's deliberate but it was convinient when I was testing.

PR #435, opened 2026-02-18
new pareto_pit function

avehtari:
#### Summary

This PR adds
- new pgeneralized_pareto() distribution function for the generalized Pareto distribution, needed by pareto_pit()
- adds tests also for the old qgeneralized_pareto() function
- adds support for weights in gpdfit(), needed by pareto_pit()
- new pareto_pit() function that behaves as pit() expect that in tails instead of using ECDF, uses CDF computed from fitted generalized Pareto distribution. This reduces variability and provides non-0 and non-1 PIT values also outside of the range of reference draws.

pareto_pit() is useful for stabilizing PIT uniformity checking, as current pit() can produce big variation whether there is single 0 or 1 or not.

The following figure illustrates how PIT values that are 0 or 1 from pit() function are replaced with a PIT value estimated with GPD fitted to the tail. Left plot is lower tail and right plot is upper tail. Log scale is used to better show the range of PIT values in tails, The dot with red circle has PIT value 0 from pit() and in log scale that is -Inf, but ggplot plots it next to the axis (and gives a warning). 

EDIT: added dashed lines showing PIT values if 1, 2, or 3 draws are further in tail than y. In the left plot one pit() result matches 3. In right plot the predicted tail probability is much smaller than 1/4000.

image

The function is specifically useful when the draws x distribution is unbounded and has nicely behaving tails, which is common for many observation families. The function is useful if we happen to have only draws from the posterior predictive distribution, but in general it would be better to compute PIT values using the known parametric observation distribution CDF.

#### Copyright and Licensing


R/pit.R
```diff
@@ -153,6 +153,254 @@ pit.rvar <- function(x, y, weights = NULL, log = FALSE, ...) {
...
+#'   \deqn{PIT(y_i) = 1 - p_{tail}(1 - F_{GPD}(y_i; c_R, \sigma_R, k_R))}
+#'
+#'   For a left-tail observation \eqn{y_i < c_L}:
+#'
+#'   \deqn{PIT(y_i) = p_{tail}(1 - F_{GPD}(-y_i; -c_L, \sigma_L, k_L))}
+#'
+#'   where \eqn{p_{tail}} is the proportion of (weighted) mass in the
+#'   tail.
+#'
+#'   When (log-)weights in in `weights` are provided, they are used for
```
florence-bockting: duplicate "in"

tests/testthat/test-gpd-distributions.R
```diff
@@ -0,0 +1,236 @@
...
+})
+
+test_that("gpdfit with weights = NULL is backward compatible", {
+  set.seed(42)
+  x <- rexp(150)
+
+  fit1 <- gpdfit(x)
+  fit2 <- gpdfit(x, weights = NULL)
+  expect_identical(fit1, fit2)
+})
```
florence-bockting: Probably add a check and corresponding error message for the case where `length(weights)` unequals `length(x)`.

````
test_that("gpdfit with length(weights) differs from length(x)", {
  set.seed(42)
  x <- rexp(200)
  w <- rep(1, 150)
  expect_error(gpdfit(x, weights = w))
})
````

Perhapse same for non-positive weight values.

florence-bockting:
I think `roxygen2::roxygenise()` wasn't running and therefore `pareto_pit.Rd` and `pgeneralized_pareto.Rd` are missing in the `man` folder (causing issues in the GitHub Actions).

Might actually be a good candidate for including this call in the GitHub Action Workflow?

PR #447, opened 2026-04-16
Add `truncate` argument to `uniformity_test()`

florence-bockting:
#### Summary

Currently, `uniformity_test()` determines automatically whether the truncated Cauchy combination test (CCT) should be used or not. This decision is based on the `test` value (POT and PRIT -> truncated CCT; PIET -> untruncated CCT).
However, there are situations in which the user should be able to change this default behaviour by customising the truncation option.

This PR adds a `truncate` argument to `uniformity_test()` which is by default `NULL` in which case it mimics the original behaviour of automatic  detection based on the `test` value.

#### Copyright and Licensing

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,9 @@
 # posterior 1.7.0
 
+### Enhancements
+
+* Add `truncate` argument to `uniformity_test()` to allow for custom truncation
+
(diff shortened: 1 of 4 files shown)


paul-buerkner (approved):
Looks good. Thank you!

PR #454, opened 2026-06-18
Assorted fixes

VisruthSK:
This PR was made with assistance from Codex.

#### Summary

Fix failures in R-devel due to changes to base `apply()`. Bumped roxygen to 8.0.0 and regenerated documentation. Also normed files to lf ending.

#### Copyright and Licensing

By submitting this pull request, the copyright holder is agreeing to license the submitted work under the following licenses:


paul-buerkner:
thank you! I have been in contact with the CRAN team and it may be that the commit that causes posterior to fail on R-devel may be reverted. But perhaps it would make them happy if posterior would still pass already now. And it may make posterior more robust against these changes in the future.

Do you expect any negative "side effects" of this PR? I.e. some aspects of rvar computation becoming more prone to bugs or errors? If not, I am happy to merge this PR.

paul-buerkner:
@mjskay do you have a few minutes to review this PR? You can judge rvar code better than I can.

VisruthSK:
I'm realizing that it would've been better to separate out the rvar changes from the other ones in this PR, but hopefully that doesn't make it too difficult to review.

I don't think there should be any major correctness or performance changes--I didn't check but I can't imagine too many people were relying on the vectrs default behaviours for as.array() and as.matrix(). I guess a rev dep check might be useful? But also might be overkill...

Also would like to get @mjskay to review if possible before merging.

PR #457, opened 2026-06-26
fix returned k when both tails give NA

avehtari:
When using `pareto_smooth(, tail = "both")` and both tails are such that NA is returned, the current version gives a warning
```
In max(left_k, right_k, na.rm = TRUE) :
  no non-missing arguments to max; returning -Inf
```
as `max()` is not able to compare NA and NA.

This PR adds a check for `left_k` and `right_k` being both NA. The new test fails without the fix and passes with the fix.

This warning came up when using `brms::loo_epred()` which calls `E_loo()` which checks both tails, and new loo v2.10.0 has switched to using `posterior` package for the checks

I used Claude to find the error and suggest a fix. I understand the fix. I wrote the new test.

### DIFF
--- a/R/pareto_smooth.R
+++ b/R/pareto_smooth.R
@@ -345,7 +345,11 @@ pareto_smooth.default <- function(x,
     )
     right_k <- smoothed$k
 
-    k <- max(left_k, right_k, na.rm = TRUE)
+    if (is.na(left_k) && is.na(right_k)) {
+      k <- NA
+    } else {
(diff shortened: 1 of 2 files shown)


florence-bockting (approved):
Looks good to me.

paul-buerkner:
I am on vacation. You can merge without my official approval

Florence Bockting ***@***.***> schrieb am Mi. 15. Juli 2026
um 15:11:

PR #463, opened 2026-09-13
Improve floating point accuracy

avehtari:
#### Summary

Herbie https://herbie.uwplse.org/ checks floating point computations, detects inaccurate expressions and finds more accurate replacements. I used Herbie with Sol to find these fixes. I have checked that all of them make sense.

- `gdp.R`: Stable upper-tail and log-probability handling
- `gdp.R`: Treat only an exactly zero GPD shape as exponential. Small nonzero shapes now retain their shape correction and finite-support behavior through the stable `log1p()` formulation.
- `pareto_smooth.R`: Stable `ps_convergence_rate()` around `k = 0.5` and `k = 1`
- `pareto_smooth.R`: Use `expm1()` for log-weight differences
- `convergence.R`: Centered fourth-moment variance calculation
- `discrete-summaries.R`: Use `log1p()` in dissent calculation

- New tests
- All new and old tests are passing

There are two new helper functions which are used once or twice, and they could be also inlined, but I think using them improves the readability.

#### Copyright and Licensing

Aki Vehtari


R/pareto_smooth.R
```diff
@@ -633,20 +635,27 @@ ps_khat_threshold <- function(ndraws, ...) {
...
-  rate[k < 0] <- 1
-  # k>0 non-finite mean
-  rate[k > 1] <- 0
-  # limit value at k=1/2
-  rate[k == 0.5] <- 1 - 1 / log(ndraws)
+  # k<=0 bounded distribution
+  rate[k <= 0] <- 1
+  # k>=1 non-finite mean
+  rate[k >= 1] <- 0
   # smooth approximation for the rest (see Appendix B of PSIS paper)
```
n-kall:
```suggestion
  # smooth approximation for the rest 
  # this is a numerically stable calculation of the equation on page 31, in Appendix B of PSIS paper
```

n-kall: As this code is now maybe less clear, I think adding a comment stating that it is the numerically stable implementation might be a good idea

R/gpd.R
```diff
@@ -57,18 +56,16 @@ pgeneralized_pareto <- function(q, mu = 0, sigma = 1, k = 0, lower.tail = TRUE,
   z <- (q - mu) / sigma
   if (abs(k) < 1e-15) {
```
n-kall: Do we still want to hardcode 1e-15 here? Maybe we can do a more sensible threshold at the same time as these changes.

n-kall:
Nice! I added one comment and one suggestion for inline comment. 
Do you have the results from herbie on how much these suggestions improve things?

avehtari:
Here is the Sol generated summary of the effect of the fixes

Expected benefit by fix

### 1. GPD tail probabilities and quantiles — high benefit

Affected code:

- `qgeneralized_pareto()`
- `pgeneralized_pareto()`

in `posterior/R/gpd.R`.

Herbie improved the upper-tail log-probability quantile kernel from **53% to 100%**.

The original implementation converted a log probability to ordinary scale and then subtracted it
from one. For an upper-tail log probability of `-100`:

```text
original quantile: Inf
correct quantile:  2425825972.048951
```

The CDF had the corresponding problem:

```text
exponential GPD at q = 100

original upper probability:     0
correct upper probability:      3.720075976020836e-44
original log upper probability: -Inf
correct log upper probability:  -100
```

Herbie scored the upper-CDF kernel at 100% even before rewriting. This demonstrates a sampling
limitation: random points rarely hit the narrow region where a representable survival probability is
lost after the lower CDF rounds to one.

The new implementation carries a log-survival probability through the calculation and only converts
to the requested output form at the end.

Expected benefits:

- Preserves extreme but representable upper-tail probabilities.
- Returns finite GPD quantiles for small upper-tail probabilities.
- Makes `lower.tail` and `log.p` behavior consistent with standard R distribution functions.
- Improves downstream Pareto-PIT calculations in extreme tails.

### 2. Pareto convergence rate — high benefit near transition points

Affected code: `ps_convergence_rate()` in `posterior/R/pareto_smooth.R`.

Herbie results:

```text
general domain:     99% → 99%
near k = 0.5:       65% → 99%
near k = 1:         70% → 98%
```

The general score concealed severe cancellation in narrow neighborhoods.

For the next representable double above `k = 0.5`, with `ndraws = 10`:

```text
original internal formula: -0.5925925925925926
original public result:     0
high-precision result:      0.6768166292078592
```

The exact `k = 0.5` branch was also discontinuous from the implemented general expression. The new
version uses algebraically equivalent `expm1()` forms selected according to the nearest singular
limit.

Expected benefits:

- Produces continuous, nonnegative rates around `k = 0.5`.
- Avoids catastrophic cancellation as `k` approaches one.
- Correctly returns one at `k = 0` and zero at `k = 1`.
- Improves diagnostic reliability precisely near the thresholds where convergence-rate values are
  most sensitive.

Away from the transition regions, the new and old formulas agreed within approximately
`7e-15` in validation sweeps.

### 3. Pareto smoothing tail differences — moderate to high benefit

Affected code: `ps_tail()` in `posterior/R/pareto_smooth.R`.

Herbie improved the focused log-weight exceedance kernel from **7% to 100%**.

The original implementation exponentiated shifted log weights and then subtracted the exponentiated
cutoff. For nearby values around `-30`:

```text
original relative error: 2.29e-4
stable relative error:   4.60e-17
```

Expected benefits:

- Preserves small positive GPD exceedances near the tail cutoff.
- Improves fitted Pareto shape and scale parameters for clustered log weights.
- Prevents avoidable zero exceedances caused by rounded exponential subtraction.
- Has little effect when tail values are well separated.

### 4. `mcse_sd()` fourth-moment calculation — high benefit for large offsets

Affected code: `mcse_sd.default()` in `posterior/R/convergence.R`.

Herbie improved the scalar fourth-moment kernel from **60% to 98%**. The implemented fix instead
addresses the vector-level source of correlated rounding error.

For 400 symmetric observations near `±1e8`:

```text
original fourth-moment difference: 5.40432e16
centered result:                   4e16

original mcse_sd: 0.0575323908112901
centered mcse_sd: 0.0494962056309299
relative error:   16.2%
```

The new calculation treats squared centered draws as a new variable and computes their variance from
centered deviations.

Expected benefits:

- Avoids cancellation between the fourth moment and squared variance.
- Prevents inaccurate or potentially negative intermediate variance estimates.
- Improves MCSE estimates for large-magnitude draws whose squared magnitudes vary only slightly.
- Leaves ordinary-scale results effectively unchanged.

### 5. Ordinal dissent logarithm — low but exact benefit

Affected code: `dissent.default()` in `posterior/R/discrete-summaries.R`.

Herbie improved the scalar kernel from **8% to 100%** by replacing

```r
log2(1 - r)
```

with the equivalent

```r
log1p(-r) / log(2)
```

For `r = 1e-20`:

```text
original result: 0
stable result:  -1.4426950408889633e-20
```

Expected benefits:

- Retains tiny nonzero category contributions.
- Improves behavior for categories located extremely close to the sample mean.
- Has negligible practical impact for most ordinal samples because larger category contributions
  dominate the total.

R/gpd.R
avehtari: Fixed: Treat only an exactly zero GPD shape as exponential. Small nonzero shapes now retain their shape correction and finite-support behavior through the stable `log1p()` formulation.

R/discrete-summaries.R
florence-bockting, line 1: This is rather a question/discussion point: 
The new helper `log2_one_minus` is currently used only for this specific case (1 caller). Thus, we could inline the code instead of creating a new function which further fragment the code base.

On the other hand, I see the benefit of having a place where we add all these helper functions related to numerical stability such that they can be reused when developing new functions. But then I would rather put them into a specific file such as `numerics.R` or similar.

R/gpd.R
```diff
@@ -20,15 +20,14 @@ qgeneralized_pareto <- function(p, mu = 0, sigma = 1, k = 0, lower.tail = TRUE,
     return(rep(NaN, length(p)))
   }
   if (log.p) {
-    p <- exp(p)
-  }
-  if (!lower.tail) {
-    p <- 1 - p
+    log_survival <- if (lower.tail) log(-expm1(p)) else p
```
florence-bockting: The current code, log(-expm1(p)), fixes the failure of log1p(-exp(p)) for probabilities near 1. There, exp(p) rounds to 1 and the result becomes -Inf. But log(-expm1(p)) fails for very small probabilities. There, 1 - exp(p) rounds to 1 and the result becomes 0. [Mächler (2012)](https://cran.r-project.org/web/packages/Rmpfr/vignettes/log1mexp-note.pdf) proposed to switch between the two formulas at p = -log(2). Each formula is used only where it is accurate.

```
# With k = 0, the generalized Pareto is the exponential distribution.

# Issues with current proposed change
##  small probability
qgeneralized_pareto(-40, k = 0, log.p = TRUE)
#> [1] 0
qexp(-40, log.p = TRUE)
#> [1] 4.248354e-18

## far upper tail
pgeneralized_pareto(40, k = 0, log.p = TRUE)
#> [1] 0
pexp(40, log.p = TRUE)
#> [1] -4.248354e-18
```

Using the idea with switching the formulas, we would have
```
log1m_exp <- function(x) {
  ifelse(x > -log(2), log(-expm1(x)), log1p(-exp(x)))
}
```
and then in the current code snippet
```
  if (log.p) {
    log_survival <- if (lower.tail) log1m_exp(p) else p
  } else {
    log_survival <- if (lower.tail) log1p(-p) else log(p)
  }
```

R/gpd.R
```diff
@@ -55,20 +54,17 @@ pgeneralized_pareto <- function(q, mu = 0, sigma = 1, k = 0, lower.tail = TRUE,
...
-  # force to [0, 1] for values outside the support
-  p <- pmin(pmax(p, 0), 1)
-  if (!lower.tail) {
-    p <- 1 - p
-  }
-  if (log.p) {
-    p <- log(p)
+  log_survival <- pmin(log_survival, 0)
+  if (lower.tail) {
+    p <- if (log.p) log(-expm1(log_survival)) else -expm1(log_survival)
```
florence-bockting: Following up on the discussion in the [previous comment](https://github.com/stan-dev/posterior/pull/463/changes#r4006217433), we would then have here:

```
p <- if (log.p) log1m_exp(log_survival) else -expm1(log_survival)
```

R/misc.R
florence-bockting, line 1: In case we decide to inline the code, we would remove these tests.

R/pareto_smooth.R
```diff
@@ -525,10 +525,12 @@ ps_tail <- function(x,
   max_tail <- max(draws_tail)
 
   if (are_log_weights) {
-    draws_tail <- exp(draws_tail)
+    tail_excesses <- exp_x_minus_exp_y(draws_tail, cutoff)
```
florence-bockting, line 528: here again the question, whether we want to have particular function for `exp_x_minus_exp_y` or rather inline code as this function has only one caller.

avehtari:
As Herbie didn't consider what happens with probabilities / densities / weights equal to 0 and log of them equal to -Inf, I added edge tests and fixed behavior for edge cases. Full NEWS

* `log_sum_exp()` shifted by `max(max(x), 0)` rather than `max(x)`, because
`warnings = FALSE` was passed to `max()`, which has no such argument and so
treated `FALSE` as an extra value to maximize over. Sums of log values below
about -745 underflowed to `-Inf`: `weights()` returned `Inf` for every draw when
given unnormalized log weights around -1000, and the normalized log weights used
by `pit()` and `pareto_pit()` were affected the same way.
* `weight_draws()` now errors when the supplied weights carry no probability
mass, instead of creating an object whose `weights()` are `NaN` and whose
`resample_draws()` fails with "missing value where TRUE/FALSE needed". The
message points at `log = TRUE`, since underflow is the usual cause. **This is a
breaking change**.
* `weights()` now errors rather than returning `NaN` when every draw has zero
weight, which can happen after subsetting an object that was valid when built.
`normalize = FALSE` still returns zeros.
* `pit()` and `pareto_pit()` no longer abort when one variable's weight column
carries no probability mass. They warn, naming the variable, and return `NA` for
it while the remaining variables are computed as usual, matching `pareto_khat()`.
* `pgeneralized_pareto()` returned `-Inf` for upper-tail log probabilities that
are perfectly representable, for example `log P(X > 50)` with `k = 0`, because it
computed the lower tail and then `1 - p`. The survival function is now carried on
the log scale throughout.
* `pgeneralized_pareto()` treated any `|k| < 1e-15` as `k = 0`. Small nonzero
shapes are now handled exactly, down to denormal `k`.
* `qgeneralized_pareto()` did not validate `p`: `q(-0.5, k = 0.2)` returned
`-0.389` and a positive log probability returned a below-support value, both
silently. Out-of-range probabilities now return `NaN` with a warning, as in the
base R quantile functions.
* `pareto_convergence_rate()` and `ps_convergence_rate()` were wrong at two exact
shape values and imprecise near one of them. A Pareto k of exactly 0 returned 0
rather than 1, so the rate jumped from 1 just below 0 to 0 at 0 and back to ~1
just above; k of exactly 0.5 dropped a factor of `S/(S-1)`, giving an error of
16% at S = 10 and 2.8e-4 at S = 4000; and the surrounding region lost about seven
digits to cancellation. The rate is now computed in a numerically stable form and
agrees with the Appendix B expression to 3e-15 across k in (0, 1).
* `pareto_convergence_rate()` and `ps_convergence_rate()` read a missing Pareto k
as 0, reporting a rate of 0, i.e. "non-finite mean", for an unknown shape, and
errored on a vector containing `NA`. Missing shapes now propagate as `NA`.
* `mcse_sd()` computed the variance of the variance as `mean(x^4) - mean(x^2)^2`,
which cancels catastrophically for location-shifted draws; it was 16% off on a
sample centred at 1e8. It is now computed in two passes.
* `dissent()` evaluated `log2(1 - x)`, which loses all precision for small `x`
and returns exactly 0 below about 1e-16.
* `ps_tail()` aborted inside `gpdfit()` with "missing value where TRUE/FALSE
needed" when the tail contained `-Inf` log weights equal to the cutoff, because
the tail-excess helper returned `NaN` for equal infinities.
* `pareto_smooth()` and `ps_tail()` compute tail excesses on the log scale, so
the generalized Pareto fit no longer loses precision when log weights in the tail
are close to the cutoff.

R/discrete-summaries.R
avehtari: I'm fine with inlining, too, but then maybe add a comment so that it's easier to see the original equation

R/gpd.R
avehtari: Ok to change. First add a failing test and then fix

florence-bockting: Ok, I created a new file misc-math.R in which I moved all functions related to numerical stability.

R/discrete-summaries.R
florence-bockting: I left it as a helper function and just moved the helper into another file: misc-math.R

R/pareto_smooth.R
florence-bockting: stayed with helper but moved it to misc-math.R

florence-bockting:
Alright, I made the changes to the file structure otherwise looks good to me.

paul-buerkner:
Thank you all! Feel free to merge when you think it's ready.

